PR #2, opened 2020-11-16
Dependency and RuboCop chores

PanisSupraOmnia:
# Summary

All that good dependency update stuff and RuboCop appeasement. Note that dropping 2.4 could technically be considered a breaking change, but it has been EOL for a little while now.

---

## Changed

- Dropped support for Ruby 2.4 (EOL) **Potentially Breaking**
- Updated dependencies
- Appeased the RuboCop


PanisSupraOmnia:
Ok, I think it's ready to go now!

Daniel-Worrall (approved):
Very brief review lookover on this, but LGTM

lib/discordrb/bot.rb
```diff
@@ -627,7 +627,7 @@ def add_await!(type, attributes = {})
       raise "You can't await an AwaitEvent!" if type == Discordrb::Events::AwaitEvent
 
       timeout = attributes[:timeout]
-      raise ArgumentError, 'Timeout must be a number > 0' if timeout&.is_a?(Numeric) && !timeout&.positive?
+      raise ArgumentError, 'Timeout must be a number > 0' if timeout.is_a?(Numeric) && !timeout&.positive?
```
swarley:
```suggestion
      raise ArgumentError, 'Timeout must be a number > 0' if timeout.is_a?(Numeric) && !timeout.positive?
```

swarley (changes requested):

lib/discordrb/bot.rb
swarley: Since we short circuit on non Numerics here, we can omit the weird safe navigation

Daniel-Worrall: On second pass, this line fails to raise as ArgumentError when passed non-Numerics (Will fail later on in the method though). Is it within scope to fix in this PR?

Daniel-Worrall: Here's a spec we can use to test it out, either in this PR or in a future PR
<details><summary>Spec</summary>

```ruby
  describe '#add_await!' do
    let(:type) { double(:type) }
    let(:timeout) { double(:timeout) }

    context 'when timeout is invalid' do
      it 'raises Argument Error when not a Numeric' do
        allow(timeout).to receive(:is_a?).with(Numeric).and_return(false)
        expect { bot.add_await!(type, timeout: timeout) }.to raise_error(ArgumentError, 'Timeout must be a number > 0')
      end

      it 'raises ArgumentError when not positive' do
        allow(timeout).to receive(:is_a?).with(Numeric).and_return(true)
        allow(timeout).to receive(:positive?).and_return(false)
        expect { bot.add_await!(type, timeout: timeout) }.to raise_error(ArgumentError, 'Timeout must be a number > 0')
      end
    end

    context 'when timeout is valid' do
      it 'does not raise' do
        allow(timeout).to receive(:is_a?).with(Numeric).and_return(true)
        allow(timeout).to receive(:positive?).and_return(true)
        expect { bot.add_await!(type, timeout: timeout) }.not_to raise_error(ArgumentError, 'Timeout must be a number > 0')
      end
    end
  end
```
</details>

swarley (approved):

PR #4, opened 2020-12-06
Create bot-status.rb

Filia-Medici-x86-64:
I made this as a example of how to change the bots status

# Summary
So i added this to the /examples to show how a bot can have custom status


swarley:
Thank you for your contribution! However, we generally don't wish to offer a full example for every functionality. In the same vein however, I may add a wiki page explaining how some actions must occur in/after `ready` since that is the main concept that isn't always intuitive.

Issue #7, opened 2020-12-20
API method review


swarley:
# Summary

While working on slash command routes, I've noticed that there are more than a few API methods with some common mistakes.

1. The final argument is not a hash, for example:
  ```ruby
    def token_webhook(webhook_token, webhook_id)
      Discordrb::API.request(
        :webhooks_wid,
        nil,
        :get,
        "#{Discordrb::API.api_base}/webhooks/#{webhook_id}/#{webhook_token}"
      )
    end
  ```
  Where the correct fix would be to add an additional `{}` as the final argument

2. Some `POST`/`PATCH` methods do not include a `content_type: :json` header entry.

3. Some methods are missing their links to the Discord documentation, particularly in `lib/discordrb/api.rb`

If anyone wants to review the methods we have and open a PR with any changes, that would be great.

PR #10, opened 2020-12-24
fix: Enforce JSON for PUT/POST requests

dali546:
# Enforce JSON for PUT/POST requests

Related to #7 . Enforce all `post/put` requests to have the `content_type: :json` header

---

## Changed
Added Content-Type header for `post/put` requests

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -97,6 +97,8 @@ module Discordrb::API
     # Specify RateLimit precision
     attributes.last[:x_ratelimit_precision] = 'millisecond'
 
+    # Force content_type to JSON if request is post/put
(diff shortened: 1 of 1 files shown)


lib/discordrb/api.rb
```diff
@@ -97,6 +97,8 @@ def request(key, major_parameter, type, *attributes)
     # Specify RateLimit precision
     attributes.last[:x_ratelimit_precision] = 'millisecond'
 
+    # Force content_type to JSON if request is post/put
+    attributes.last[:content_type] = :json if %i[put post].include?(type)
```
swarley, line 101: Unfortunately we cannot do this, as certain webhook and message based requests require the use of `multipart/form-data`.

swarley (changes requested):

PR #22, opened 2021-01-09
Remove extra attempted heartbeat after stopping Bot normally

kaine119:
# Summary

Changes the sequence of time delays in the heartbeat thread.

If the delay happens after checking the session, the session could be
suspended while the thread is sleeping, causing us to try to heartbeat on a
suspended session. Moving the sleep down sends the heartbeat right after the
check.

---

## Changed
- `Gateway`'s `@heartbeat_thread`: Moved the heartbeat_interval delay to after raising the heartbeat event, before
checking if the session is suspended.

### DIFF
--- a/lib/discordrb/gateway.rb
+++ b/lib/discordrb/gateway.rb
@@ -444,13 +444,16 @@ module Discordrb
       @heartbeat_interval = interval
       @heartbeat_thread = Thread.new do
         Thread.current[:discordrb_name] = 'heartbeat'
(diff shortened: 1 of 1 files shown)


swarley:
We need to do a few more things for this change. Ideally we would kill the heartbeat loop on disconnect, and allow it to restart on HELLO, so we also aren't trying to send a heartbeat without an active connection.

We also need to kill the thread that detects zombie connections, since that will continue to attempt to reconnect as well.

PR #28, opened 2021-02-03
Update webhooks gemspec to match

rorhug:
When the both webhooks and the main library were updated to 3.4.0, the gemspec pointing to webhooks from main was left on 3.3.0.

3.3.0 used `rest-client >= 2.1.0.rc1` which was reverted here https://github.com/shardlab/discordrb/commit/648596335da886b01ed87ff6f43c8bc2a7e0aae4

This causes incompatibilities when using rest-client 2.0.x in other gems.


swarley:
Thanks!

PR #33, opened 2021-02-05
Add reaction awaits to Message

### LINKED ISSUE #30: Feature Request: Message#add_reaction_await(!)
To add `Message#add_reaction_await` and `Message#add_reaction_await!`

I feel like adding an await for a reaction is common enough to warrant a handy abstraction for it.

Currently, `Messsage#add_await` adds a `MessageEvent` await, and I thought we could extend this to have it parse a symbol (like `:message` or `:reaction_add`) for the type of event (and default to message), then allow for optional arguments like emoji.

Not sure if this is supported in the current await system, but it would be nice to support multiple emoji for the reaction awaits.

kaine119:
# Summary
New Message methods that add an await for a `ReactionAddEvent` for that message.
Fixes #30 

---

## Added
- `Message#await_reaction`, `Message#await_reaction!`

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -222,6 +222,19 @@ module Discordrb
       @bot.add_await!(Discordrb::Events::MessageEvent, { from: @author.id, in: @channel.id }.merge(attributes), &block)
     end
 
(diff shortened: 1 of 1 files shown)


swarley:
Thank you!

PR #36, opened 2021-03-01
Fix test for Ruby 3.0

mataku:
# Summary

Fix test for positional arguments and keyword arguments separation for Ruby 3.0.

See more about keyword arguments behaviour: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

## How to reproduce

```shell
# Use ruby 3.0
rbenv local 3.0.0 

gem i bundler
bundle install

# You'll see the test failing on main branch.
bundle exec rspec
```
## Fixed

`spec/commands_spec.rb`


swarley:
Thank you!

PR #40, opened 2021-03-07
Use query string builder instead of string interpolation for some queries

kaine119:
# Summary
Use `URI.encode_www_form` to build URI queries for multiple parameters instead of the current string interpolation.

---

## Changed
`API::Channel`, `API::Server`


swarley:
Thanks!

PR #50, opened 2021-05-14
Update documentation for Message#reply!

biqqles:
# Summary

#3 added support for inline replies, which is fantastic, but it seems the documentation for `Message#reply!` was accidentally copied over from `Channel#send_message`. This patch updates that documentation.

I debated putting _inline reply_ in quotation marks since as far as I can find it's not an official term in [the API](https://discord.com/developers/docs/resources/channel#message-object) (an inline reply is simply referred to as "a reply"), but decided it only made the sentence more confusing.

---

## Fixed

Incorrect description for `Message#reply!`

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -158,7 +158,7 @@ module Discordrb
       @channel.send_message(content)
     end
 
-    # Sends a message to this channel.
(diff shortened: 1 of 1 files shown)


PanisSupraOmnia (approved):
LGTM

swarley:
Thank you!

PR #53, opened 2021-05-21
Check if we have the SERVER_MEMBERS intent before fetching server members

### LINKED ISSUE #51: Server#members hangs when bot doesn't use SERVER_MEMBERS intent
# Summary
Running a bot without the SERVER_MEMBERS intent will cause Server#members to hang. 

Debug output looks like this: 
```
[3] pry(main)> bot.servers[TEST_SERVER_ID].members
[DEBUG : main @ 2021-05-15 17:34:18.890] Members for server <TEST_SERVER_ID> not chunked yet - initiating
[OUT : main @ 2021-05-15 17:34:18.890] {"op":8,"d":{"guild_id":<TEST_SERVER_ID>,"query":"","limit":0}}
```

---

## Environment
**Ruby version:**
`ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin18]`

**Discordrb version:**
`/Users/me/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/discordrb-359326e8743c`

kaine119:
# Summary
Fixes #51.

The gateway doesn't respond if we try to ask for server members without first declaring the intent. Originally, `Server#members` would just wait for a response that isn't coming, causing it to never return. This change checks the intents we sent with the gateway first before requesting members.

---

## Added
- Exposed `Gateway#intents`, the final calculated intents parameter sent to the gateway
- `Server#members` now raises a RuntimeError if the gateway was not initialized with the SERVER_MEMBERS intent

### DIFF
--- a/lib/discordrb/data/server.rb
+++ b/lib/discordrb/data/server.rb
@@ -143,10 +143,15 @@ module Discordrb
     end
 
     # @return [Array<Member>] an array of all the members on this server.
+    # @raise [RuntimeError] if the bot was not started with the :server_member intent
     def members
       return @members.values if @chunked
 
       @bot.debug("Members for server #{@id} not chunked yet - initiating")
+
+      # If the SERVER_MEMBERS intent flag isn't set, the gateway won't respond when we ask for members.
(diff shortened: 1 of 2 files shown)


swarley:
Thanks!

PR #54, opened 2021-05-22
Fix YARD for Message#reply

### LINKED ISSUE #47: Deprecate alias Channel#send
Object#send is used by RubyMine debugger, and we shouldn't be overwriting any Object methods.

Also, turns out [I'm to blame](https://github.com/shardlab/discordrb/commit/b5e19dfc95491b4078c8c1b32142645d0852bf14#diff-f2f2d924d0a1e92b90bd01aeaf25d31286afc910ea5b97a98d269431112d939bR1048)

biqqles:
# Summary

~~Deprecates `Channel#send` (fixes #47) for reasons described in that issue.~~ I also took the opportunity to trivially fix the generated documentation for `Message#reply` (also deprecated - previously it had the wrong signature) and DRY the `@return` of `Message#reply!`. 

---

## Fixed
Incorrect YARD signature for `Message#reply`


swarley:
Thanks!

PR #55, opened 2021-05-23
Allow Bot#initialize to request only unprivileged intents

biqqles:
# Summary

With v8, intents are enforced when identifying. By default the library will request all intents, including privileged ones, even if the bot does not use them. This PR makes it easier for the authors of such bots to only request unprivileged intents as opposed to all intents.

It also allows you to specify _no_ intents. Previously the documentation suggested that you could use `nil` for this, but this would have caused an error.

---

## Added
`Discordrb::UNPRIVILEGED_INTENTS`
`Discordrb::NO_INTENTS`

## Changed
`Bot#initialize`: `intents` can now be equal to `:unprivileged`
`Bot#initialize`: `intents` can now be equal to `:none`


swarley:
Thanks!

PR #58, opened 2021-05-24
Allow CommandBot to take nil intents

biqqles:
# Summary

`Bot#initialize` is allowed to take `nil` for its `intents` keyword argument. However, `CommandBot` breaks this by replacing `nil` with `:all`.

---

## Fixed
`CommandBot#initialize` can now take `intents: nil` like its superclass


biqqles:
Oops, will fix this in my other PR since it's easier.

PR #61, opened 2021-06-01
Fixup: Paginate reactions

Daniel-Worrall:
Fixes sending a limit over 100 as this is an error for the API requests
The Paginator object handles limits.
The API method handles limit defaults.

# Summary

When you send Message#reacted_with with a limit over 100, it will pass this to the API method and fail.

---

## Fixed
Message#reacted_with now accepts limits over 100


Daniel-Worrall:
Turns out I can't change the source of a PR, so this'll just stay as is until reviewed and merged

swarley:
Thanks!

PR #70, opened 2021-08-13
Require message_id for Channel#load_message

mattantonelli:
# Summary

Ensures the `message_id` parameter of `Channel#load_message` is not nil since we cannot load the message without it. Calling the method with a nil value will now return a more helpful `ArgumentError` instead of a `Discordrb::Errors::UnknownError`.

## Added

Added a check to `Channel#load_message` to ensure the `message_id` parameter is not nil.

### Before
```rb
bot.channel(12345).message(nil)

[ERROR : main @ 2021-08-13 17:40:53.860] 404: Not Found
Discordrb::Errors::UnknownError: 404: Not Found
```

### After
```rb
bot.channel(12345).message(nil)

ArgumentError: message_id cannot be nil
```


swarley:
Thanks!

PR #75, opened 2021-09-01
Fix typo in slash commands

CryroFox:
I thought I already made a PR for this lol

# Summary
It's just a typo fix, thats all this is.
I only noticed this using this example file to get an idea of how this api works

---

## Added
+ Singular letter into a comment
## Changed
Slash_Commands.rb
## Fixed
A singular typo :P


swarley:
Thanks!

PR #78, opened 2021-09-09
Fix Interaction#button

### LINKED ISSUE #77: Interaction#button is unusable
# Summary

The `Interaction#button` method attempts to return the button that triggered the interaction based on the value of `@type`. However, this value will always be type 3: component, because it is the **interaction type**, not the component type. As a result, the method will always return `nil`.

```rb
# @return [Hash, nil] Returns the button that triggered this interaction if applicable, otherwise nil
def button
  return unless @type == TYPES[:button]

  Components::Button.new(@data, @bot)
end
```

Additionally, various information about the button is unavailable because the button is being constructed from the top level `data` of the Interaction response. Values such as `label` and `style` are only accessible via the `message.components`, as shown [here](https://discord.com/developers/docs/interactions/message-components#component-interaction-object-sample-component-interaction).

I believe both issues can be resolved by restoring the `Interaction#button` method from my original [pull request](https://github.com/shardlab/discordrb/pull/62/files#diff-d6dfe0b90efeac0adae59abffeb76820fddef93bcc4e86081f0e0b74071176a1R191-R200) to add button support.

---

**Discordrb version:**

3.4.2 87ae658

mattantonelli:
# Summary

Fixes #77

## Fixed

- Fixed the type check to look for the component interaction type
- Returning a button constructed within an `ActionRow` instead of constructing it from the top-level data. This is necessary because the top-level data does not contain button-specific fields like label and style.
- Corrected the documented type of `Interaction#data`

### DIFF
--- a/lib/discordrb/data/interaction.rb
+++ b/lib/discordrb/data/interaction.rb
@@ -47,7 +47,7 @@ module Discordrb
     # @see TYPES
     attr_reader :type
 
-    # @return Hash The interaction data.
+    # @return [Hash] The interaction data.
     attr_reader :data
 
     # @!visibility private
--- a/lib/discordrb/data/interaction.rb
+++ b/lib/discordrb/data/interaction.rb
@@ -240,9 +240,13 @@ module Discordrb
 
     # @return [Hash, nil] Returns the button that triggered this interaction if applicable, otherwise nil
     def button
-      return unless @type == TYPES[:button]
+      return unless @type == TYPES[:component]
 
-      Components::Button.new(@data, @bot)
+      @message['components'].each do |row|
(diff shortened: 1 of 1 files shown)


lib/discordrb/data/interaction.rb
```diff
@@ -47,7 +47,7 @@ class Interaction
     # @see TYPES
     attr_reader :type
 
-    # @return [String] The interaction data.
+    # @return Hash The interaction data.
```
swarley:
```suggestion
    # @return [Hash] The interaction data.
```

swarley (changes requested):

swarley (approved):

PR #86, opened 2021-09-19
Deprecate group DM endpoint methods

### LINKED ISSUE #16: Test/Remove group channel methods?
https://github.com/shardlab/discordrb/blob/main/lib/discordrb/api/channel.rb#L334-L400

These API methods were added by myself and at least one of them will only work with a user token.

They should be revisited and tested to comply with the up to date API

kaine119:
# Summary
Bots aren't allowed in group DMs, and while the [API docs](https://discord.com/developers/docs/resources/channel#group-dm-add-recipient) list a bunch of methods for managing group DMs, they aren't meant for bots to use, and don't work as currently implemented.

The API methods are meant to be used by the (deprecated) GameBridge SDK to invite people to a temporary "room" on Discord, relying on each party authenticating with an OAuth application separately.

Fixes #16 

---

## Deprecated
`Discordrb::API::Channel`
* `.create_empty_group`
* `.create_group`
* `.add_group_user`
* `.remove_group_user`
* `.leave_group`

### DIFF
--- a/lib/discordrb/api/channel.rb
+++ b/lib/discordrb/api/channel.rb
@@ -332,6 +332,7 @@ module Discordrb::API::Channel
   end
 
   # Create an empty group channel.
+  # @deprecated Discord no longer supports bots in group DMs, this endpoint was repurposed and no longer works as implemented here.
   # https://discord.com/developers/docs/resources/user#create-group-dm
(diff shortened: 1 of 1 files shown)


swarley:
Thanks!

PR #94, opened 2021-10-03
Split ServerEmojiEvent out from ServerEvent class

blerg-rush:
# Summary

In investigating the refactor required to re-enable the [Lint/MissingSuper](https://github.com/shardlab/discordrb/blob/7c101bcac52424ed930c91fbf399d73b096ee740/.rubocop.yml#L17) cop, I found `ServerEmojiEvent` effectively inherits nothing from `ServerEvent` besides its parent class.

If we want to _usefully_ call the initializers of our parent classes, `ServerEvent` and `ServerEmojiEvent` child classes behave differently enough (e.g., `ServerEvent` usually attempts to set its `server` attribute based on the provided `data` hash's `id` value while `ServerEmojiEvent` requires `server` as an argument;  the `data` argument(s) for the server and emoji events have different meanings) that it makes more sense for `ServerEmojiEvent` to be an event class of its own.

I'd like to follow this up with the remaining `Lint/MissingSuper` fixes so that cop can be re-enabled, but this was a large enough change to stand on its own.

---

## Added
* Added `ServerEmojiEventHandler` to provide `#matches?` for each server emoji event handler class

## Changed
* Changed `ServerEmojiEvent`'s parent class from `ServerEvent` to `Event`
* Refactored `ServerEmojiEvent` child classes to call their parent class's initializer and inherit its `bot` and `server` attribute readers


blerg-rush:
Updated at @swarley's recommendation; in the off chance someone depends on `ServerEmojiEvent` to subclass `ServerEvent` in their event handling logic, removing that relationship could break their code.

PR #102, opened 2021-11-06
refactor: Use chunk_index and chunk_count to determine chunk status.

swarley:
# Summary

This changes chunking calculation to use the discord provided `chunk_index` and `chunk_count` fields instead of checking for expected length. This should help us avoid some difficult to trace deadlocks.

---

## Changed
`Server#process_chunk` - now accepts an additional two arguments; `chunk_index` and `chunk_count`.

### DIFF
--- a/lib/discordrb/data/server.rb
+++ b/lib/discordrb/data/server.rb
@@ -87,7 +87,6 @@ module Discordrb
 
       # Whether this server's members have been chunked (resolved using op 8 and GUILD_MEMBERS_CHUNK) yet
       @chunked = false
-      @processed_chunk_members = 0
 
       @booster_count = data['premium_subscription_count'] || 0
       @boost_level = data['premium_tier']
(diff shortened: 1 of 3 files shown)


lib/discordrb/data/server.rb
```diff
@@ -805,19 +804,16 @@ def invites
     # Processes a GUILD_MEMBERS_CHUNK packet, specifically the members field
     # @note For internal use only
     # @!visibility private
-    def process_chunk(members)
+    def process_chunk(members, chunk_index, chunk_count)
       process_members(members)
-      @processed_chunk_members += members.length
-      LOGGER.debug("Processed one chunk on server #{@id} - length #{members.length}")
+      LOGGER.debug("Processed chunk #{chunk_index + 1}/#{chunk_count} server #{@id} - length #{members.length}")
```
Daniel-Worrall: I think we should include the raw chunk_index in debug logs, to stay consistent with the API.

lib/discordrb/data/server.rb
```diff
@@ -805,19 +804,16 @@ def invites
...
 
-      # Don't bother with the rest of the method if it's not truly the last packet
-      return unless @processed_chunk_members == @member_count
+      return unless chunk_index + 1 == chunk_count
```
Daniel-Worrall: Probably will never matter in a bot runtime but I'd prefer this be `chunk_index < chunk_count`

Daniel-Worrall (changes requested):
Lgtm, just minor QoL things

PR #104, opened 2021-11-14
fix: force guild recache from GUILD_CREATE

swarley:
# Summary

Adds an option to `ensure_server` to force the recache of a guild. This is necessary because there is a window in this cache lifecycle where we can have a server object with bad state that permanently occupies the cache.

image

## Added
- `Cache#ensure_server` - Add `force_recache` parameter to ensure guilds from `GUILD_CREATE` are used as the source of truth.

## Changed
- `Server#update_data` - Move data initialization logic from `initialize` to `update_data` for reusability.

### DIFF
--- a/lib/discordrb/cache.rb
+++ b/lib/discordrb/cache.rb
@@ -147,10 +147,14 @@ module Discordrb
 
     # Ensures a given server object is cached and if not, cache it from the given data hash.
     # @param data [Hash] A data hash representing a server.
+    # @param force_cache [true, false] Whether the object in cache should be updated with the given
+    #   data if it already exists.
     # @return [Server] the server represented by the data hash.
-    def ensure_server(data)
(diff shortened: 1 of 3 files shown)


lib/discordrb/cache.rb
```diff
@@ -147,10 +147,14 @@ def ensure_user(data)
 
     # Ensures a given server object is cached and if not, cache it from the given data hash.
     # @param data [Hash] A data hash representing a server.
+    # @param force_cache [true, false] Whether the object in cache should be updated with the given
```
PanisSupraOmnia, line 150: The `Boolean` placeholder type is accepted yard convention so we could probably use that here.

PanisSupraOmnia (commented):
The actual logic LGTM, just one docs nitpick from me.

Daniel-Worrall (approved):

PR #107, opened 2021-11-28
refactor!: Servers are now referred to as guilds

swarley:
# Summary

Too many occurrences to count. All instances where `Server` was used to refer to a `Guild` object have been replaced to correctly use guild. This includes models, methods, and events.


swarley:
Moving this into the icebox. This change will be further discussed in DAPI and decided on later in the development cycle of 4.0

PR #111, opened 2021-12-02
Allow bulk overwriting application commands

TeeSeal:
# Summary

Implemented method for bulk overwriting application commands and added a couple more builders in the context of interactions to facilitate that. The `ApplicationCommandBuilder` is a bit awkward because I didn't want to make any breaking changes to the existing interface of `Bot#register_application_command` and `Bot#edit_application_command`. 

---

## Added

- `Interactions::ApplicationCommandBuilder` - needed something to encapsulate all data contained in an application command to facilitate the creation of the `ApplicationCommandSetBuilder` (see below). 
- `Interactions::ApplicationCommandSetBuilder` - builder that exposes simple DSL to define multiple application commands within a single block
- `Bot#bulk_overwrite_application_commands` - public method that allows... well... bulk overwriting application commands. Works as shown in the YARD example.
```ruby
bot.bulk_overwrite_application_commands(server_id: ENV["TEST_SERVER_ID"]) do |app|
  app.command(:ping, "Pong!")
  app.command(:echo, "Echo some text") do |cmd|
    cmd.string(:text, "Text to echo")
  end
end
```

## Changed

- `API::Application.create_global_command`
- `API::Application.edit_global_command`
- `API::Application.create_guild_command`
- `API::Application.edit_guild_command`

All 4 of the above have been updated to take a hash of command parameters instead of separate arguments.
```ruby
{
  name: :test,
  description: "some command",
  options: [],
  default_permission: true,
  type: 1
}
```
This could pose problems with `.create_global_command` and `.create_guild_command` as the name and description for those are no longer *techincally* required. Suggestions on handling this are welcome.

### DIFF
--- a/lib/discordrb/api/application.rb
+++ b/lib/discordrb/api/application.rb
@@ -44,13 +44,13 @@ module Discordrb::API::Application
 
   # Edit a global application command.
   # https://discord.com/developers/docs/interactions/slash-commands#edit-global-application-command
-  def edit_global_command(token, application_id, command_id, name = nil, description = nil, options = nil, default_permission = nil, type = 1)
+  def edit_global_command(token, application_id, command_id, command_data)
     Discordrb::API.request(
       :applications_aid_commands_cid,
       nil,
       :patch,
       "#{Discordrb::API.api_base}/applications/#{application_id}/commands/#{command_id}",
-      { name: name, description: description, options: options, default_permission: default_permission, type: type }.compact.to_json,
+      command_data.to_json,
       Authorization: token,
       content_type: :json
     )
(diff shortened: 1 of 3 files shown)


swarley:
I think this looks good for the most part but I'm not a fan of removing the parameters in favor of a single hash for the API level methods. That's a blocker for getting this merged. If we can get that reverted and some eyes from another contributor on this we should be good to go

swarley:
I know it's been a while since you've touched this (my apologies), I'm more than happy to make some changes myself to push this through if you've moved on from the feature 🙂

TeeSeal:
@swarley yeah it has been a minute 😅 I no longer need this myself but I don't mind looking into it again and making the changes. See what I can do tomorrow 😃

TeeSeal:
@swarley I've updated the PR to not touch existing code outside of a [small fix](https://github.com/shardlab/discordrb/pull/111/files#diff-20ecafab3dff76e2f494acaf73563a1df0350835b08ed6861d5787302d09a732R879) and documentation.

TeeSeal:
Been cleaning up my PRs. This is too old and probably no longer relevant. Closing.

PR #120, opened 2022-01-13
fix: Use retry-after response header if missing in the response body for 429 API bans

Daniel-Worrall:
# Summary

In the case of a 429 after an API ban, the retry-after in the JSON body isn't present, but is present in the headers.

The library will incorrectly wait for 0.0s since it's `nil` in the body and spam the API. This fixes this issue.

## Fixed
Use retry-after response header if retry-after is missing in the response body

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -143,7 +143,7 @@ module Discordrb::API
 
       unless mutex.locked?
         response = JSON.parse(e.response)
-        wait_seconds = response['retry_after'].to_f
+        wait_seconds = response['retry_after'] ? response['retry_after'].to_f : e.response.headers[:retry_after].to_i
         Discordrb::LOGGER.ratelimit("Locking RL mutex (key: #{key}) for #{wait_seconds} seconds due to Discord rate limiting")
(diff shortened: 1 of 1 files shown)


sadleb:
Related PR for reference: https://github.com/shardlab/discordrb/pull/120

We just had this re-surface today and continue to dig into the root cause of the 429 ban in the first place.

sadleb:
> Related PR for reference:
> 
> * [fix: 429 runaway retry. `retry_after` is in seconds and so is `sleep` #114](https://github.com/shardlab/discordrb/pull/114)
> 
> We just had this re-surface today and continue to dig into the root cause of the 429 ban in the first place.

fyi: Apparently it's b/c we shared an IP address with others and either they were making too many requests and got rate-limited or the aggregate of all folks sharing that IP led to rate-limiting for that IP. This app is on Heroku.

swarley:
The root cause of the reported issue was due to a known problem with Discord and hosts that have shared IPs. This fix is still a good one because it's possible that Discord will send a 429 that does not include the more precise value in the body. Thanks @Daniel-Worrall! 🚀

PR #146, opened 2022-03-28
feat: add server avatar

### LINKED ISSUE #145: Per server avatars

swarley:
# Summary

Add server avatar support.

Closes #145

---

## Added
- `MemberAttributes#avatar_id`
- `Member#avatar`

## Changed
- `MemberAttributes#avatar_id` no longer delegates to `User` directly
- `Member#avatar` no longer delegated to `User` directly

### DIFF
--- a/lib/discordrb/data/member.rb
+++ b/lib/discordrb/data/member.rb
@@ -24,8 +24,8 @@ module Discordrb
     alias_method :timeout, :communication_disabled_until
 
     # @!attribute [r] avatar_id
-    #   @return [String, nil] the ID of this user's current avatar, can be used to generate an avatar URL.
-    #   @see User#avatar_url
+    #   @return [String, nil] the ID of this member's current avatar, can be used to generate an avatar URL.
+    #   @see Member#avatar_url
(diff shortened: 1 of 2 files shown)


lib/discordrb/data/member.rb
```diff
@@ -294,6 +302,13 @@ def display_name
       nickname || username
     end
 
+    # (See User#avatar_url)
```
Daniel-Worrall, line 305:
```suggestion
    @see User#avatar_url
```

lib/discordrb/data/member.rb
```diff
@@ -294,6 +302,13 @@ def display_name
...
+    def avatar_url(format = nil)
```
Daniel-Worrall, line 306: Include a parameter to get the global avatar_url, regardless of a Member's avatar_id, since the Member doesn't provide easy access to the underlying `@user`, and this overwrites one of the methods.

Also, would be more appropriate on MemberAttributes for consistency (#avatar_url is on UserAttributes) before the v4 rework

PR #147, opened 2022-04-10
Image Size Parameter

### LINKED ISSUE #131: Support size parameter for images
To support `User#avatar_url` (and by extension `API::User.avatar_url`) size passing.

There will be other similar methods that return an image URL that should also support sizes.

PanisSupraOmnia:
# Summary

Closes #131.

This PR adds the ability to specify a size for the image returned for a variety of CDN endpoints. I also improved the Overcommit configuration since I happened to run into an issue with it while committing the changes for this PR; I realize these are unrelated changes so I can split them to a separate PR if you'd like.

---

## Added

- The size of image can now be requested for CDN-served images.

## Fixed

- Made Overcommit invoke RuboCop via Bundler, to ensure we are using the currently specified version.
- Re-enabled Overcommit's `AuthorName` check because the original behavior of the check that prompted it to be disabled has been changed upstream; first and last name are no longer required, anything of non-zero length is sufficient.


swarley:
I'm fine with size being a kwarg here, but for consistency's sake I wonder if we should keep it as an optional positional. I'd like some input

Issue #136, opened 2022-02-21
CircleCI Deprecating 14.04/16.04 Machine Images


PanisSupraOmnia:
Not specifying a machine image defaults to using a Trusty image, [which is deprecated and will soon fail.](https://circleci.com/blog/ubuntu-14-16-image-deprecation/) Two options to fix: specify the latest machine image or use a Docker image instead.

PR #151, opened 2022-04-23
Add new channel types.

connorshea:
# Summary
There are three channel types not currently represented in the gem, although I guess it may not be worth adding these values if we don't support their corresponding functionality in the gem?

Also, not sure if forum (15) should be included since it's in active development.

See https://discord.com/developers/docs/resources/channel#channel-object-channel-types

---

## Added

- Added the newest channel types to the TYPES hash in Discordrb::Channel:
  - `GUILD_STAGE_VOICE` as `stage_voice`
  - `GUILD_DIRECTORY` as `directory`
  - `GUILD_FORUM` as `forum`

### DIFF
--- a/lib/discordrb/data/channel.rb
+++ b/lib/discordrb/data/channel.rb
@@ -20,6 +20,9 @@ module Discordrb
       news_thread: 10,
       public_thread: 11,
       private_thread: 12
+      stage_voice: 13,
+      directory: 14,
+      forum: 15
     }.freeze
 
     # @return [String] this channel's name.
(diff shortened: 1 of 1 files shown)


connorshea:
Also just noticed we still have the `store` type but that's not mentioned in the Discord API documentation anymore 🤔

swarley:
Weird that CI passed because you're missing a comma from the previous list 😄, if you can fix that we should be good to go

Daniel-Worrall:
CI didn't run at all is why. Broken config = no checks = pass :^)

Daniel-Worrall:
Discussed in dAPI, #136 is most likely the cause of the CI. Will push a fix soon

connorshea:
> Weird that CI passed because you're missing a comma from the previous list 😄, if you can fix that we should be good to go

Too much work, back to potato farming

swarley (approved):

swarley:
Thanks!

PR #155, opened 2022-05-01
Add to_message to Interactions::Message

dylhack:
# Summary

This would create a shortcut to using all of the methods that are available for Discordrb::Message

## Added

  - Added Interactions::Message#to_message (2c87b416a89efd2c705fbe20bd81128fe22dfa6d)
  - Added to_message to Discordrb::Message (0970ff27c2a70b26e0092af3fafff107651aae86)
  
 Adding to_message to Discordrb::Message was in inspiration of Ruby's String#to_s method.

### DIFF
--- a/lib/discordrb/data/interaction.rb
+++ b/lib/discordrb/data/interaction.rb
@@ -725,7 +725,7 @@ module Discordrb
         Discordrb::Message.new(@data, @bot)
       end
 
-      alias_method :to_message, :message
+      alias_method :message, :to_message
 
       # @!visibility private
       def inspect
--- a/lib/discordrb/data/interaction.rb
+++ b/lib/discordrb/data/interaction.rb
@@ -642,6 +642,7 @@ module Discordrb
 
       # @!visibility private
       def initialize(data, bot, interaction)
+        @data = data
         @bot = bot
         @interaction = interaction
         @content = data['content']
@@ -718,6 +719,13 @@ module Discordrb
       def edit(content: nil, embeds: nil, allowed_mentions: nil, components: nil, &block)
         @interaction.edit_message(@id, content: content, embeds: embeds, allowed_mentions: allowed_mentions, components: components, &block)
       end
+
+      # @return [Discordrb::Message]
+      def to_message
+        Discordrb::Message.new(@data, @bot)
+      end
+
+      alias_method :to_m, :to_message
 
       # @!visibility private
       def inspect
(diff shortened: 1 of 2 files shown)


dylhack:
note: The hash of the data parameter in Interactions::Message is missing `referenced_message`.

lib/discordrb/data/interaction.rb
```diff
@@ -719,6 +720,13 @@ def edit(content: nil, embeds: nil, allowed_mentions: nil, components: nil, &blo
         @interaction.edit_message(@id, content: content, embeds: embeds, allowed_mentions: allowed_mentions, components: components, &block)
       end
 
+      # @return [Discordrb::Message]
+      def to_message
+        Discordrb::Message.new(@data, @bot)
+      end
+
+      alias_method :to_m, :to_message
```
Daniel-Worrall: This alias doesn't go with the rest of our codebase, we don't use any `to_x` single letter aliases beyond the stdlib stuff

lib/discordrb/data/message.rb
```diff
@@ -390,5 +390,13 @@ def buttons
...
       results.flatten.compact
     end
+
+    # to_message -> self or message
+    # @return [Discordrb::Message]
+    def to_message
+      self
+    end
+
+    alias_method :to_m, :to_message
```
Daniel-Worrall: ditto

Daniel-Worrall (changes requested):
Please remove the aliases, they don't fit with the rest of the library (consistency is very important)

Otherwise, the code seems fine, though it comes into question why store the data as well as everything it contains.

This should probably be reworked to contain a message object if possible, and use delegated properties instead of bringing everything to the Interaction object itself.
Not a fault of the PR, just a design decision we may not want to lock ourselves into that I only just caught since I haven't had any chance to look at Interactions

Opinions, @swarley?

swarley:
I think the data storage is fine, we want to be lazy here. I'm not a fan of the `to_m`. I would prefer just `#message` because I believe it's the most consistent

lib/discordrb/data/message.rb
dylhack: resolved in 93f13b52d1b471bdd1f1f9313f50dc62fa4e8df4

lib/discordrb/data/interaction.rb
dylhack: resolved in 93f13b52d1b471bdd1f1f9313f50dc62fa4e8df4

Daniel-Worrall (approved):
LGTM 👍

dylhack:
Awesome!

swarley:
Thank you!

Issue #175, opened 2022-08-24
Wrong rate_limit_per_user interval for Channel


Dakurei:
For some unknown reason I was banned from the development Discord server, so I come to make the request directly here ...

The interval does not correspond at all to the official documentation, preventing to set a cooldown of more than 2 minutes.
https://github.com/shardlab/discordrb/blob/dc27fe18463da3ccfd0f0266030aa7ad51b2c2b9/lib/discordrb/data/channel.rb#L350

The value should be **21600** instead of **120**
![screenshot](https://user-images.githubusercontent.com/8146474/186430602-3f410eff-c325-49e7-96b1-7b88759c33e2.png)

Thanks in advance for the modification!

swarley:
More than happy to update this

@Dakurei can you join our server and give me your discord ID so I can look into that for you? https://discord.gg/Mz2hm8Vw

swarley:
Nevermind, I found your case and unbanned you from the Discord API server 🙂

Dakurei:
Thanks a lot !

Dakurei:
I made the modification, and I took the occasion to fix the test that was failing.

Normally it should be good, I'll let you check and merge it when you have the opportunity 👍

PR #176, opened 2022-08-24
Modification of the maximum value of rate_limit_per_user

Dakurei:
In order to match the official API, the maximum value is increased from 120 to 21600

Issue #175


swarley (approved):
Thanks!

PR #178, opened 2022-10-02
Support multipart file uploads for interaction responses via webhook builder

robherley:
# Summary

Currently, if you wish to [attach a file to an interaction](https://discord.com/developers/docs/reference#uploading-files) via the webhook builder it does not work, ie:

```ruby
file = File.open('/Users/robherley/Pictures/photo.png')

event.respond(content: 'photo') do
  builder.file = file
end
```

The above will only send the content, and not the file as intended.

This is because the methods defined in the webhook API don't support a `file` parameter and the interaction data is only serialized for a json payload:

https://github.com/shardlab/discordrb/blob/dc27fe18463da3ccfd0f0266030aa7ad51b2c2b9/lib/discordrb/data/interaction.rb#L97-L100

This PR adds support for file uploads via the interaction methods that support it. To avoid breaking changes and compatibility with the current method signature, I added an additional nil-by-default param. However, it would be nice to move these all to keyword arguments eventually.

I also added some specs for the modules that were missing them. They don't have complete coverage of the module, but cover the changes introduced in this PR.

---

## Added

- `Discordrb::Webhooks::Builder#to_payload_hash`: returns relevant hash depending on the presence of `@file` 
- spec for `Discordrb::API::Interaction`
- spec for `Discordrb::Interaction`

## Changed

- `Discordrb::API::Interaction`
  - added `file` as a parameter to `#create_interaction_response`
- `Discordrb::API::Webhook`
  - added `file` as a parameter to `#token_edit_message`
- `Discordrb::Interaction`
  - for `#respond`, `#update_message`, `#edit_response`, `#send_message`, `#edit_message`:
    - using `#to_payload_hash` instead of `#to_json_hash` for webhook builder
    - passing `file` param from webhook builder to relevant API method

## Deprecated

None

## Removed

None

## Fixed

- Uploading file in interaction methods via webhook builder
- The unrelated failing `Channel#delete_all_emoji_reactions` test, it was a small mistake with the emoji encoding

### DIFF
--- a/lib/discordrb/api/interaction.rb
+++ b/lib/discordrb/api/interaction.rb
@@ -32,8 +32,8 @@ module Discordrb::API::Interaction
 
   # Edit the original response to an interaction.
   # https://discord.com/developers/docs/interactions/slash-commands#edit-original-interaction-response
-  def edit_original_interaction_response(interaction_token, application_id, content = nil, embeds = nil, allowed_mentions = nil, components = nil)
-    Discordrb::API::Webhook.token_edit_message(interaction_token, application_id, '@original', content, embeds, allowed_mentions, components)
+  def edit_original_interaction_response(interaction_token, application_id, content = nil, embeds = nil, allowed_mentions = nil, components = nil, file = nil)
+    Discordrb::API::Webhook.token_edit_message(interaction_token, application_id, '@original', content, embeds, allowed_mentions, components, file)
   end
 
   # Delete the original response to an interaction.
--- a/lib/discordrb/api/interaction.rb
+++ b/lib/discordrb/api/interaction.rb
@@ -6,16 +6,21 @@ module Discordrb::API::Interaction
 
   # Respond to an interaction.
   # https://discord.com/developers/docs/interactions/slash-commands#create-interaction-response
-  def create_interaction_response(interaction_token, interaction_id, type, content = nil, tts = nil, embeds = nil, allowed_mentions = nil, flags = nil, components = nil)
+  def create_interaction_response(interaction_token, interaction_id, type, content = nil, tts = nil, embeds = nil, allowed_mentions = nil, flags = nil, components = nil, file = nil)
     data = { tts: tts, content: content, embeds: embeds, allowed_mentions: allowed_mentions, flags: flags, components: components }.compact
+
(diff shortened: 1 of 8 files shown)


robherley:
@swarley is this PR good to merge as is? or would you prefer support for [`attachments`](https://discord.com/developers/docs/resources/channel#attachment-object) vs. [the existing single file implementation](https://github.com/shardlab/discordrb/blob/dc27fe18463da3ccfd0f0266030aa7ad51b2c2b9/lib/discordrb/webhooks/builder.rb#L36-L43)

swarley:
@robherley If you're up to the task we can get the multi file upload going here, otherwise I can take care of expanding it in a separate PR. I'd like to get it working with multiple files before we tag a new release 🙂. Let me know what you want to do

robherley:
@swarley I can add the multi-file attachments in this PR! Just a couple impl questions:
- would you prefer removing `file` from the webhook builder altogether? or should I just mark it as deprecated and append the single file to `attachments`
- a lot of the `::Interaction` methods use positional args instead of keyword args which makes invocation a bit awkward [like this](https://github.com/shardlab/discordrb/blob/ad88821f05e4c217a6589256cfbcb81899363729/spec/api/interaction_spec.rb#L43-L53). are you opposed to changing those func signatures to use keyword arguments instead?

swarley:
I would mark as deprecated, and I'd like to change the API methods to use kwargs in the next major release but currently they need to stay as positional for consistency and semver reasons.
I'm actively working on the API portion of v4 now and rest assured we won't have positional stacks like this

robherley:
@swarley multi-file attachments should be supported now:

```ruby
fileA = File.open('/Users/robherley/Pictures/photoA.png')
fileB = File.open('/Users/robherley/Pictures/photoB.png')

event.respond(content: 'photos!') do |builder|
  builder.attachments = [fileA, fileB]
end
```

`file` will log out as deprecated and if a user supplies _both_ it just falls back to the deprecated behavior until they switch completely to `attachments`. I thought about appending the `file` to the `attachments` list but that'll probably just create more confusing behavior 😅 

also noticed the rubocop complaining about unchanged files 🤷 I can fix those if you want

PR #180, opened 2022-10-03
Add support for slash command mentions

uscoders82:
# Summary

This PR aims to allow `.mention` on an ApplicationCommand
---
~Added the method `.mention` with the alias `.to_s`

## Added
~The ability to mention slash commands


swarley (approved):
Thanks!

PR #184, opened 2022-10-11
Permissions Flags errors

Dakurei:
# Summary

Hi !

There is a big problem with the permission flags listing. Some bits are missing and the last ones are offset because of this.

I could see this by doing an overwrite that was supposed to give me the use_private_threads permission, but it gives me use_public_threads when looking at the permissions on Discord.

Current permissions flags in Discordrb main:
image

Actual permissions flags on Discord API:
image

That's why I directly made a pull request, in order not to depend on a monkeypatch too long before it's merge.

Thanks in advance !

---

## Changed
- Change lines inside differents files to be Rubycop friendly

## Fixed
- Fix Discordrb::Permissions::FLAGS values
- Fix /spec/api/channel_spec.rb broken again


spec/api/channel_spec.rb
```diff
@@ -55,7 +55,7 @@
 
   describe '.delete_all_emoji_reactions' do
     let(:message_id) { double('message_id', to_s: 'message_id') }
-    let(:emoji) { "\u{1F525}" }
+    let(:emoji) { '\u{1F525}' }
```
swarley, line 58: This is not correct, double quotes are necessary for the unicode format

swarley:
Closed in favor of #209

PR #189, opened 2022-10-26
Fix button modal

Dakurei:
# Summary

The 1st problem was that opening the modal from a button inside simple message was sending the Hash of the message data in the function instead of a Message object directly, so the .components method was unavailable.

The 2nd problem (which at first sight didn't bother the function, but returned an exception in the logs) was that the show_modal request was sent again during the modal_submit, which didn't make sense.

(Based on PR #184)

---

## Fixed

- Instantiate Discordrb::Message in the method Discordrb::Interaction.get_component to prevent crash when Hash of message data was passed instead of Discordrb::Message directly
- Make sure that the show_modal request is not sent back unnecessarily during a modal_submit to avoid log flood


swarley:
Closed in favor of #209

PR #191, opened 2022-10-29
Fix rubocop offenses from modals

uscoders82:
# Summary

This PR fixes all the rubocop offenses caused by the new modals feature added to discordrb
---

## Fixed
~ Use ENV.fetch('DISCORDRB_TOKEN') instead of ENV['DISCORDRB_TOKEN']
~ Pass &:components as an argument to flat_map instead of a block

### DIFF
--- a/examples/modals.rb
+++ b/examples/modals.rb
@@ -3,9 +3,9 @@
 require 'discordrb'
 require 'securerandom'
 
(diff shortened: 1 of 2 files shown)


Nekka-Lucifer:
As raised by Devonte in discord, this change would set the server id to the bot token.

uscoders82:
Oops, I'm not sure how I didn't notice that but I think I fixed it now!

PR #192, opened 2022-11-15
Add missings components

Dakurei:
# Summary

This pull request aims to make available, in Discordrb, the new "Select" components that are defined in the official Discord API.

image

These look like this on the client

image

(Based on PR #189)

---

## Added
- Adding components: user_select, role_select, mentionable_select and channel_select
- Adding new events and event handlers to separate the types of data received easily (and not receive everything in a single event, unless it makes sense to add such event)
> UserSelectEvent, UserSelectEventHandler, RoleSelectEvent, RoleSelectEventHandler, MentionableSelectEvent, MentionableSelectEventHandler, ChannelSelectEvent and ChannelSelectEventHandler
- Adding example file to test this new components

## Changed
- Changing component select_menu to string_select (for consistency with the official API), but an alias is defined with the old name to avoid introducing a breaking change
- Changing class name SelectMenuEvent and SelectMenuEventHandler to StringSelectEvent and StringSelectEventHandler
- Changing SelectMenuBuilder initializer to add a select_type argument in order to reuse this builder for new components


Dakurei:
Force push to change the modification concerning the initializer of SelectMenuBuilder, the previous modification would have caused a breaking change for people who used the SelectMenuBuilder object directly in their code.

swarley:
Closed in favor of #209

PR #193, opened 2022-11-30
Allow access to the entire banlist of a server

Dakurei:
# Summary

Having a big server, with a large banlist, I noticed when I wanted to export it, that Discordrb didn't have the necessary arguments in its request to the API to display more than the first 1000 IDs.

Current request:
image
_https://drb.shardlab.dev/main/Discordrb/API/Server.html#bans-class_method_

Against what should be done according to the API:
image
_https://discord.com/developers/docs/resources/guild#get-guild-bans_

That's why I added the necessary arguments inspired by Channel#history.

(Based on PR #192)

---

## Changed
- Added the necessary arguments to the Server#bans method (and by extension Discordrb::API::Server#bans) to allow access to the entire banlist and not just the first 1000 by default.


swarley:
Closed in favor of #209

PR #196, opened 2022-12-27
fix: Server#max_emoji returning wrong value

Birdie0:
# Summary

`Server#max_emoji` has typo in it by referring to `@level` instead of `@boost_level` attribute, hence it returns `50` every time.

## Fixed

- Replaced `@level` with correct `@boost_level`


swarley (approved):
Thanks!

PR #198, opened 2022-12-29
Feat: Forums

franciscofpll:
# Summary
This pull request adds support for channels with forum type.

## Added
The ForumType(15) already exists in the channel file, so I just add news tests and the DirectoryType(14) and ForumType(15) on the Data::Interaction class.

### DIFF
--- a/lib/discordrb/data/interaction.rb
+++ b/lib/discordrb/data/interaction.rb
@@ -415,7 +415,10 @@ module Discordrb
         news_thread: 10,
         public_thread: 11,
         private_thread: 12,
-        stage: 13
+        stage: 13,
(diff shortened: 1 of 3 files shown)


Droid00000 (changes requested):
Good PR so far, but to get this across the finish line, forum tags, and the default reaction object would need to be added along with a few other things. This includes an update to the `Server#create_channel` method to allow for type 15 as a valid channel type. 

There seem to have been many updates and additions to forum channels since this PR was submitted, so it may be a good idea to take a quick peek at the API docs again.

Droid00000:
Forums were properly implemented in #400.

PR #199, opened 2022-12-29
Fixes rubocop and specs

PixeLInc:
# Summary

This pr fixes the long broken rubocop, along with the failing specs.
So we are now green bois again.

---

## Fixed

- examples/modals.rb to use `ENV.fetch` instead of `ENV[...]`
- lib/discordrb/data/interaction.rb to prefer proc rather than a block to please lord rubocop.
- spec/data/webhook_spec.rb failing due to missing `hash_including` call..
- spec/api/channel_spec.rb failing due to double quotes around the unicode emoji

### DIFF
--- a/examples/modals.rb
+++ b/examples/modals.rb
@@ -3,9 +3,9 @@
 require 'discordrb'
 require 'securerandom'
 
-bot = Discordrb::Bot.new(token: ENV['DISCORDRB_TOKEN'])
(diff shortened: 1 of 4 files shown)


Dakurei:
Fixing the webhook problem does not require changing the test, but fixing the bug in the class directly. I already did it in one of my last PR

https://github.com/shardlab/discordrb/pull/192/commits/591d1277d265d5d57db2fa77ae311f792044d8a5

PR #204, opened 2023-03-06
fix: make Message#author lazy

### LINKED ISSUE #117: Error: Unknown Member in `Channel#history`
## Description

When calling `Channel#history` with a limit larger than the amount of messages that exist in the channel, it would seem like the method is attempting to get author details for `nil` messages, and the result is this error appearing multiple times:

```
[ERROR : main @ 2022-01-10 11:00:39.235] Unknown Member
[ERROR : main @ 2022-01-10 11:00:39.511] Unknown Member
[ERROR : main @ 2022-01-10 11:00:39.803] Unknown Member
# ... Ctrl+C is required to abort
# and each line takes some 200ms to complete, hinting a web request is in progress.
```

## Reproduction

Run the below code on a channel that has less than 100 messages.

```ruby
require 'discordrb'
channel_id = 123 # replace me with a channel ID
bot = Discordrb::Bot.new token: ENV['BOT_TOKEN']
channel = bot.channel channel_id
pp channel.history 100
```

## Environment

**Ruby version:**

```
ruby 3.0.1p64 (2021-04-05 revision 0fb782ee38) [x86_64-linux]
```

**Discordrb version:**

```
  * discordrb (3.4.2 a1083b0)
  * discordrb-webhooks (3.4.2 a1083b0)
```

biqqles:
# Summary
Fixes #117. Calling `Message#author` will still result in 'Unknown Member' being printed if the member does not exist (don't think there's a way to hide this? It is kinda annoying), but it won't hang instantiators of `Message` such as `Channel#history`.

This is a bit rough right now but functional, will make a cleanup pass when I have more time.

---

## Fixed
- Reimplemented `Message#author` as a method instead of an attribute

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -10,12 +10,6 @@ module Discordrb
     alias_method :text, :content
     alias_method :to_s, :content
 
-    # @return [Member, User] the user that sent this message. (Will be a {Member} most of the time, it should only be a
-    #   {User} for old messages when the author has left the server since then)
-    attr_reader :author
-    alias_method :user, :author
-    alias_method :writer, :author
-
     # @return [Channel] the channel in which this message was sent.
     attr_reader :channel
 
@@ -88,34 +82,7 @@ module Discordrb
 
       @server = @channel.server
 
-      @author = if data['author']
-                  if data['author']['discriminator'] == ZERO_DISCRIM
-                    # This is a webhook user! It would be pointless to try to resolve a member here, so we just create
-                    # a User and return that instead.
-                    Discordrb::LOGGER.debug("Webhook user: #{data['author']['id']}")
-                    User.new(data['author'], @bot)
-                  elsif @channel.private?
-                    # Turn the message user into a recipient - we can't use the channel recipient
-                    # directly because the bot may also send messages to the channel
-                    Recipient.new(bot.user(data['author']['id'].to_i), @channel, bot)
-                  else
-                    member = @channel.server.member(data['author']['id'].to_i)
-
-                    if member
-                      member.update_data(data['member']) if data['member']
-                    else
-                      Discordrb::LOGGER.debug("Member with ID #{data['author']['id']} not cached (possibly left the server).")
-                      member = if data['member']
-                                 member_data = data['author'].merge(data['member'])
-                                 Member.new(member_data, @server, bot)
-                               else
-                                 @bot.ensure_user(data['author'])
-                               end
-                    end
-
-                    member
-                  end
-                end
+      @author_data = data['author']
 
       @webhook_id = data['webhook_id'].to_i if data['webhook_id']
 
@@ -352,7 +319,7 @@ module Discordrb
 
     # The inspect method is overwritten to give more useful output
     def inspect
-      "<Message content=\"#{@content}\" id=#{@id} timestamp=#{@timestamp} author=#{@author} channel=#{@channel}>"
+      "<Message content=\"#{@content}\" id=#{@id} timestamp=#{@timestamp} author=#{author} channel=#{@channel}>"
     end
 
     # @return [String] a URL that a user can use to navigate to this message in the client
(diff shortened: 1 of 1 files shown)


lib/discordrb/data/message.rb
```diff
@@ -88,34 +82,7 @@ def initialize(data, bot)
...
-                    User.new(data['author'], @bot)
-                  elsif @channel.private?
-                    # Turn the message user into a recipient - we can't use the channel recipient
-                    # directly because the bot may also send messages to the channel
-                    Recipient.new(bot.user(data['author']['id'].to_i), @channel, bot)
-                  else
-                    member = @channel.server.member(data['author']['id'].to_i)
-
-                    if member
-                      member.update_data(data['member']) if data['member']
```
biqqles, line 109: From a very quick check, `data['member']` doesn't seem to exist anymore so I ignored the code relating to it.

lib/discordrb/data/message.rb
```diff
@@ -390,5 +357,36 @@ def buttons
...
+        # Turn the message user into a recipient - we can't use the channel recipient
+        # directly because the bot may also send messages to the channel
+        Recipient.new(@bot.user(author_id), @channel, @bot)
+      else
+        author = @channel.server.member(author_id) || @bot.user(author_id)
+        case author
+        when Member
+          author.update_data(@author_data)
+        when User
+          Discordrb::LOGGER.debug("Member with ID #{author_id} not available (probably left the server).")
```
swarley:
```suggestion
          Discordrb::LOGGER.debug("Member with ID #{author_id} not available (possibly left the server).")
```

lib/discordrb/data/message.rb
swarley: `data['member']` is present when messages are received via a gateway event

lib/discordrb/data/message.rb
```diff
@@ -390,5 +357,36 @@ def buttons
 
       results.flatten.compact
     end
+
+    # The user who sent this message message. Will be a {Member} most of the time, a {User} in the case
+    # that the user has left the server, or a {Recipient} if the message was sent in a private channel.
```
swarley: Please add a type doc here

lib/discordrb/data/message.rb
```diff
@@ -390,5 +357,36 @@ def buttons
...
+        User.new(data['author'], @bot)
+      elsif @channel.private?
+        # Turn the message user into a recipient - we can't use the channel recipient
+        # directly because the bot may also send messages to the channel
+        Recipient.new(@bot.user(author_id), @channel, @bot)
+      else
+        author = @channel.server.member(author_id) || @bot.user(author_id)
+        case author
+        when Member
+          author.update_data(@author_data)
```
swarley: Because we're handling this lazily now we should not update at this point in the lifecycle. There is no guarantee that our data is more up to date than the cache

swarley (changes requested):

lib/discordrb/data/message.rb
swarley: We may need to ensure that we're updating/creating the member during the MessageCreate event processing though

biqqles: Is this what `ensure_user` is for?

lib/discordrb/data/message.rb
```diff
@@ -390,5 +357,32 @@ def buttons
 
       results.flatten.compact
     end
+
+    # The user who sent this message message. Will be a {Member} most of the time, a {User} in the case
+    # that the user has left the server, or a {Recipient} if the message was sent in a private channel.
+    # @return [Member, User, Recipient]
+    def author
+      author_id = @author_data['id'].to_i
+      return unless author_id
```
biqqles, line 384: This and L378 mean return value can also be nil which is not in type doc. Not sure of the circumstances when this could happen, perhaps for a deleted account?

swarley: In the case of a deleted account there will still be a dummy user associated with it. I'm not aware of a message that can exist without an author

biqqles: Do you want me to remove the guard then?

swarley: I think we're fine to leave it after some thought

lib/discordrb/data/message.rb
```diff
@@ -408,5 +375,32 @@ def to_message
...
+
+      if @author_data['discriminator'] == ZERO_DISCRIM
+        # This is a webhook user! It would be pointless to try to resolve a member here, so we just create
+        # a User and return that instead.
+        LOGGER.debug("Webhook user: #{author_id}")
+        User.new(data['author'], @bot)
+      elsif @channel.private?
+        # Turn the message user into a recipient - we can't use the channel recipient
+        # directly because the bot may also send messages to the channel
```
swarley, line 393: We should use `@bot.ensure_user` here if we've been passed the user data

lib/discordrb/data/message.rb
```diff
@@ -408,5 +375,32 @@ def to_message
...
+        Recipient.new(@bot.user(author_id), @channel, @bot)
+      else
+        author = @channel.server.member(author_id) || @bot.user(author_id)
```
swarley: Similarly, we should use ensure_member/ensure_user

swarley (commented):
Last two requests on this one in hopes of avoiding more network requests if possible

lib/discordrb/data/message.rb
biqqles: `member` is cached by default but made the other changes

Droid00000:
Hey! Thank you for this work! Really sorry that we missed this and it went stale. Superseded by #130

PR #205, opened 2023-03-09
feat: optimize user caching for invites

swarley:
## Summary

Use `ensure_user` to avoid making an API call during invite creation.


Daniel-Worrall (approved):
lgtm

PR #194, opened 2022-11-30
Support multiple embeds

Dakurei:
# Summary

Fixes the problem preventing the creation and editing of messages with multiple embeds

(Based on PR #193)

---

## Fixed
Fix #132

PR #209, opened 2023-03-24
Fix broken ip discovery

### LINKED ISSUE #208: Voice is broken
# Summary

IP Discovery is broken
---
IP Discovery is broken due to the breaking change that Discord released on March 20th. When I try to call `.voice_connect` it fails to establish a connection, but it worked perfectly fine before the change.
## Environment

NetBSD 9.3

**Ruby version:**

ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [x86_64-netbsd]
**Discordrb version:**
  * discordrb (3.4.2)
  * discordrb-webhooks (3.4.2)
  * ffi (1.15.5)
  * opus-ruby (1.0.1)
  * rest-client (2.1.0)
  * websocket (1.2.9)
  * websocket-client-simple (0.6.1)

Dakurei:
# Summary

Fixes the problem of the change in IP Discovery

(Based on PR #194)

---

## Fixed
Fix #208


lib/discordrb/data/server.rb
```diff
@@ -585,9 +585,13 @@ def max_emoji
...
     end
 
+    # Retrieve banned users from this server.
+    # @param limit [Integer] Number of users to return (up to maximum 1000, default 1000).
+    # @param before_id [Integer] Consider only users before given user id.
+    # @param after_id [Integer] Consider only users after given user id.
     # @return [Array<ServerBan>] a list of banned users on this server and the reason they were banned.
-    def bans
-      response = JSON.parse(API::Server.bans(@bot.token, @id))
+    def bans(limit = nil, before_id = nil, after_id = nil)
```
swarley: Would prefer to see these as kwargs

lib/discordrb/voice/network.rb
```diff
@@ -98,10 +98,21 @@ def send_audio(buf, sequence, time)
...
+      # Add Type request (0x1 = request, 0x2 = response)
+      discovery_packet += [0x1].pack('n')
+
+      # Add Length (excluding Type and itself = 70)
+      discovery_packet += [70].pack('n')
+
+      # Add SSRC
+      discovery_packet += [@ssrc].pack('N')
+
+      # Add 66 zeroes so the packet is 74 bytes long
```
swarley, line 113: More efficient way to handle it
```suggestion
     # request type, packet length and SSRC padded to 70 bytes
      discovery_packet = [0x1, 70, @ssrc].pack('nnNx66')
```

lib/discordrb/voice/network.rb
```diff
@@ -98,10 +98,21 @@ def send_audio(buf, sequence, time)
...
       discovery_packet += "\0" * 66
```
swarley, line 114: This line would be removed if using the suggestion above

swarley (commented):
Looks good, small notes and I'd like feedback from some other contributors on whether or not they agree with my opinion that the select events should inherit a common base event type. @Daniel-Worrall @PanisSupraOmnia

lib/discordrb/data/server.rb
Dakurei: Should I assume that the whole chain of PRs is concerned to be accepted?
Just to do like for the emoji test and add it as a commit at the end of the chain on this branch, or do you want me to do it properly in the concerned branch and rebase and force push all the following ones?

lib/discordrb/voice/network.rb
Dakurei: I totally agree that it is more efficient. I had hesitated to do it this way, but then I thought that if Discord changed the packet again, it would be easier for anyone to come back and add what was missing from the packet by just looking at the Discord doc like I did.
If you want me to reduce the packet building process to the bare minimum though, I can do that.

lib/discordrb/data/server.rb
swarley: We can work off this PR and merge it all at once

lib/discordrb/voice/network.rb
swarley: We can leave it this way. More readable is probably better than minor efficiency anyway

lib/discordrb/data/server.rb
Dakurei: Ok I'll do that then!

lib/discordrb/voice/network.rb
Dakurei: In theory the call to this function is only made at the connection in a voice room, unless you have a bot that spends its time doing this, there is little chance that non-optimization would have any negative impact.

Then I can see to compact the 3 first ones (type, length and ssrc) by commenting well at least, and by leaving the padding "\0" apart ?

swarley: I think we should leave it exactly as you have it. If we switch to a complicated pack call we end up making code that relatively few people can change in the future. I was a bit over zealous in optimization 🙂

swarley (approved):

swarley:
Thanks a lot! I really appreciate the effort that went into all these changes. Sorry that it took so long to get around to the review

Dakurei:
No worries, I did the best I could, and in the meantime most of my bots were running on a branch of my fork with all these changes.

It's easier if it's directly on the real "main" branch, but I was in no hurry personally 🤣

PR #211, opened 2023-03-27
Fix `.make_localization` not giving ArgumentError

uscoders82:
# Summary

Make .make_localization return ArgumentError instead of giving an error that doesn't help the user.
---
I messed up a part of the localization PR, and the error was raised only if nil was passed to both name and description, but the method defaulted to `''` instead of `nil`. This fixes that.

## Fixed
ArgumentError not being raised when no fields are given for `.make_localization`


Droid00000:
I think the fix this was for ended up being reverted entirely

PR #213, opened 2023-04-04
fix an error with webhooks view initialize

max-fry-apps:
[ERROR : it-1092744109200003092 @ 2023-04-04 12:34:58.465] Exception: #<ArgumentError: wrong number of arguments (given 7, expected 1..6)>
[ERROR : it-1092744109200003092 @ 2023-04-04 12:34:58.465] .../discordrb/lib/discordrb/webhooks/view.rb:125:in `initialize'
[ERROR : it-1092744109200003092 @ 2023-04-04 12:34:58.465] .../discordrb/lib/discordrb/webhooks/view.rb:67:in `new'
[ERROR : it-1092744109200003092 @ 2023-04-04 12:34:58.465] .../discordrb/lib/discordrb/webhooks/view.rb:67:in `string_select'

PR #214, opened 2023-04-04
Bugfix - Kwarg forgotten in string_select method

Dakurei:
# Summary

Kwarg forgotten in string_select method probably during the last rebases.
And adding disabled attributes that were not put on the new components during the last rebases.

Fix PR #213

---

## Added
- Adding disabled attributes that were not put on the new components during the last rebases
## Fixed
- Kwarg forgotten in string_select method


swarley (approved):
Thanks!

PR #221, opened 2023-05-24
New usernames system

Dakurei:
# Summary

I added the new 'global_name' field that would be interesting to have for bots, as well as modified 2-3 mechanics of some methods to keep some kind of consistency with what Discord is displaying since this transition period with the new usernames.

I didn't touch anything concerning the updating of this field from the bot, I only did the reading.

I voluntarily left some TODOs lying around since I'm not sure of anything about the direction the API modification is going to take, and I had to search in the github of their docs (https://github.com/discord/discord-api-docs/pull/6130) since the deployed one is still not up to date concerning this new system...

I also tried to get rid of the dependency on the 0000 discriminator to determine if a user is a webhook

And I added a missing flag while I was at it.

---

## Added
- Add global_name attribute to User
- Add User#webhook_account (with aliases: webhook_account? & webhook?)
- Add ACTIVE_DEVELOPER flag
## Changed
- Change User#distinct to use discriminator or not, depend to the context
- Change Discordrb::API::User#default_avatar & User#avatar_url to use old method to calculate index or not, depend to legacy or not (Depending on whether the account has already migrated to the new username)
- Change Member#display_name to select global_name before username, but after server nickname
- Change Webhook#avatar_url to use new method to calculate index (After testing the webhook creation, the ID already seems to be used for the default avatar instead of the discriminator at 0000)
- Change Message#initialize & User#initialize to add a way to know if the user is a webhook without relying on the discriminator
## Removed
- Remove Message::ZERO_DISCRIM which is no longer used


tarragon:
Any chance this can go in soon?

swarley (approved):

Issue #185, opened 2022-10-12
Allow kicking user from voice channels with `Server.move` by passing `nil` for `channel_id`


gregdan3:
Discord.js and discordpy both support this feature by passing their equivalent of `nil`:

- https://discordpy.readthedocs.io/en/latest/api.html#discord.Member.move_to
- https://discord.js.org/#/docs/discord.js/main/class/VoiceState?scrollTo=setChannel

discordrb's equivalent, `Server.move`, evaluates the channel's ID and fails if `nil` is passed.
- https://drb.shardlab.dev/v3.4.2/Discordrb/Server.html#move-instance_method

Admittedly, I don't immediately see how this is possible via the API, as the modify user voice state endpoint doesn't allow for a null channel id.

https://discord.com/developers/docs/resources/guild#modify-current-user-voice-state

Still, given this is implemented in two other libraries, it must be possible.

Sobsz:
Comfusingly, the relevant request is actually [Modify Guild Member](https://discord.com/developers/docs/resources/guild#modify-guild-member), as Modify User Voice State is stated to only work on stage channels. The former's documentation states:
> If the `channel_id` is set to null, this will force the target user to be disconnected from voice.

PR #226, opened 2023-06-07
Allow disconnect user from a voice channel via the Server#move method

Dakurei:
# Summary

It's all in the title
Allows to fix issue #185 

---

## Changed
- Server#move accept nil for channel param

### DIFF
--- a/lib/discordrb/data/server.rb
+++ b/lib/discordrb/data/server.rb
@@ -619,8 +619,8 @@ module Discordrb
       API::Server.remove_member(@bot.token, @id, user.resolve_id, reason)
     end
 
-    # Forcibly moves a user into a different voice channel. Only works if the bot has the permission needed.
-    # Can't move user if he's not already connected to a server voice channel.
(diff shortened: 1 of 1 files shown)


lib/discordrb/data/server.rb
```diff
@@ -620,10 +620,11 @@ def kick(user, reason = nil)
     end
 
     # Forcibly moves a user into a different voice channel. Only works if the bot has the permission needed.
+    # Can't move user if he's not already connected to a server voice channel.
```
anna328p: May be better to rephrase this, merging it with the other requirement for this call and avoiding generic pronouns. e.g.

```ruby
# Forcibly moves a user into a different voice channel.
# Only works if the bot has the permission needed and if the user is already connected to some voice channel on this server.
```

anna328p (changes requested):
otherwise LGTM

anna328p (approved):

swarley (approved):

PR #230, opened 2023-06-27
Add tags to role model

gracz21:
# Summary

I added a wrapper class for the Discord role's tags and added the tags field to the role model as it wasn't possible to access the role tags data (e.g. role's bot ID) while using the server role created/updated/deleted gateway event payload.

---

## Added
- added `Discordrb::Role::Tags` class

## Changed
- added `Discordrb::Role#tags` attribute

### DIFF
--- a/lib/discordrb/data/role.rb
+++ b/lib/discordrb/data/role.rb
@@ -60,7 +60,7 @@ module Discordrb
 
       def initialize(data)
         @bot_id = data['bot_id']&.resolve_id
-        @integration_id = data['integration_id']
+        @integration_id = data['integration_id']&.resolve_id
         @premium_subscriber = data.key?('premium_subscriber')
         @subscription_listing_id = data['subscription_listing_id']
         @available_for_purchase = data.key?('available_for_purchase')
--- a/lib/discordrb/data/role.rb
+++ b/lib/discordrb/data/role.rb
@@ -59,7 +59,7 @@ module Discordrb
       attr_reader :guild_connections
 
       def initialize(data)
-        @bot_id = data['bot_id']
+        @bot_id = data['bot_id']&.resolve_id
         @integration_id = data['integration_id']
         @premium_subscriber = data.key?('premium_subscriber')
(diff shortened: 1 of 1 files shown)


lib/discordrb/data/role.rb
```diff
@@ -35,6 +35,39 @@ class Role
...
+      attr_reader :subscription_listing_id
+
+      # @return [true, false] Whether this role is available for purchase
+      attr_reader :available_for_purchase
+
+      # @return [true, false] Whether this role is a guild's linked role
+      attr_reader :guild_connections
+
+      def initialize(data)
+        @bot_id = data['bot_id']
```
Dakurei:
```suggestion
        @bot_id = data['bot_id']&.resolve_id
```
To be consistent with your documentation, which indicates that you return Integer

lib/discordrb/data/role.rb
```diff
@@ -35,6 +35,39 @@ class Role
...
+        @integration_id = data['integration_id']
```
Dakurei:
```suggestion
        @integration_id = data['integration_id']&.resolve_id
```
To be consistent with your documentation, which indicates that you return Integer

lib/discordrb/data/role.rb
```diff
@@ -35,6 +35,39 @@ class Role
...
+        @premium_subscriber = data.key?('premium_subscriber')
+        @subscription_listing_id = data['subscription_listing_id']
```
Dakurei:
```suggestion
        @subscription_listing_id = data['subscription_listing_id']&.resolve_id
```
To be consistent with your documentation, which indicates that you return Integer

Dakurei (approved):
Error in returned type according to documentation (String instead of Integer)

gracz21:
Thanks for review @Dakurei. All suggestions were applied

swarley (approved):
Looks good, thank you both.

gracz21:
Cool, thanks

PR #232, opened 2023-07-15
Fix: Catching send_temporary_message exceptions

arminherling:
# Summary

I was trying to send an error message if my bot couldnt send a temporary message in a different channel and found a small bug in send_temporary_message.
Its currently not possible to catch exceptions because send_temporary_message creates the message in a different thread before deleting it after the timeout.

```rb
    begin
      different_channel.send_temporary_message content, 30
    rescue Discordrb::Errors::NoPermission => e
      next "No Permission!: #{e}"       # This branch never runs because we cant catch the exception currently
    end
```

## Fixed

We can now catch the exception in user code


arminherling:
Would be nice if someone could review this PR

PR #233, opened 2023-07-22
feature: add new intents

Birdie0:
# Summary

Added missing intents, but mainly specifying message content intent as privileged one.

---

## Added

Added next intents to INTENTS array:

- `MESSAGE_CONTENT (1 << 15)`
- `GUILD_SCHEDULED_EVENTS (1 << 16)`
- `AUTO_MODERATION_CONFIGURATION (1 << 20)`
- `AUTO_MODERATION_EXECUTION (1 << 21)`

## Changed

Added `MESSAGE_CONTENT` intent to exclude it from list of unprivileged ones

### DIFF
--- a/lib/discordrb.rb
+++ b/lib/discordrb.rb
@@ -33,10 +33,12 @@ module Discordrb
     direct_messages: 1 << 12,
     direct_message_reactions: 1 << 13,
     direct_message_typing: 1 << 14,
-    message_content: 1 << 15,
+    server_message_content: 1 << 15,
     server_scheduled_events: 1 << 16,
     auto_moderation_configuration: 1 << 20,
-    auto_moderation_execution: 1 << 21
+    auto_moderation_execution: 1 << 21,
+    server_message_polls: 1 << 24,
+    direct_message_polls: 1 << 25
   }.freeze
 
   # All available intents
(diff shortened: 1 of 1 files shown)


lib/discordrb.rb
```diff
@@ -32,15 +32,19 @@ module Discordrb
     server_message_typing: 1 << 11,
     direct_messages: 1 << 12,
     direct_message_reactions: 1 << 13,
-    direct_message_typing: 1 << 14
+    direct_message_typing: 1 << 14,
+    message_content: 1 << 15,
```
Droid00000: I think maybe `server_message_content` may be a better name since the intent only effects message content in servers. DMs along with a [few other circumstances](https://discord.com/developers/docs/topics/gateway#message-content-intent) are unaffected.

lib/discordrb.rb
```diff
@@ -32,15 +32,19 @@ module Discordrb
...
+    server_scheduled_events: 1 << 16,
+    auto_moderation_configuration: 1 << 20,
+    auto_moderation_execution: 1 << 21
```
Droid00000: If it's not too much trouble, could you please add the following new intents as well: 
`server_message_polls: 1 << 24`
`direct_message_polls: 1 << 25`

Droid00000 (changes requested):
Just needs a few more new intents, and a minor wording change!

Birdie0:
all done!

Droid00000 (approved):
Thanks!

Droid00000:
After thinking about it a little, I think it's better to just let the relevant feature PRs update the intents hash, e.g. #365, #334, and #35

PR #235, opened 2023-09-13
Adds the ability to define a custom status for a bot

Dakurei:
# Summary

Possible for some time apparently (https://discord.com/developers/docs/change-log#aug-8-2023)
However, a bot can only define customizable text, no emoji can be defined (With the exception of putting a unicode emoji directly into the status text, only server emojis cannot be used in this way, of course 😎)

image

+ Adds a test/example file for custom status display

---

## Added
- Method `Bot#custom_status=` to set custom status
- Add a test/example file `examples/custom_status.rb`

## Changed
- Method `Bot::update_status‎` to add custom status


lib/discordrb/bot.rb
```diff
@@ -548,7 +548,11 @@ def update_status(status, activity, url, since = 0, afk = false, activity_type =
       @streamurl = url
       type = url ? 1 : activity_type
 
-      activity_obj = activity || url ? { 'name' => activity, 'url' => url, 'type' => type } : nil
+      activity_obj = if activity_type == 4
+                       { 'name' => activity, 'type' => type, 'state' => activity }
```
tarragon, line 560: `type` here will (or should) always be 4. If something is changed in the logic above and `type` ends up not 4 then this might cause unwanted behaviour.

Setting `'type' => 4` or `'type' => activity_type` might be better.

tarragon:
Tested this locally and it worked as expected, looks good to me!

lib/discordrb/bot.rb
Dakurei: The only problem I can see is that I used `activity_type == 4` instead of `type == 4` by mistake, which could have caused inconsistency in some cases.

But there's no reason to hard-code the value instead of using that of the variable, as long as it's the right one that's used and verified by the condition

I force push the change

PR #240, opened 2023-10-14
Allow supressing embeds on Message

Quintasan:
# Summary

* Add `pry-byebug` to development dependencies
* Add `API::Channel#suppress_embeds`
* Add `Message#suppress_embeds`

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -73,6 +73,9 @@ module Discordrb
     # @return [Array<Component>]
     attr_reader :components
 
+    # @return [Integer] flags set on the message
+    attr_reader :flags
+
     # @!visibility private
     def initialize(data, bot)
       @bot = bot
@@ -157,6 +160,7 @@ module Discordrb
 
       @components = []
       @components = data['components'].map { |component_data| Components.from_data(component_data, @bot) } if data['components']
+      @flags = data['flags'] || 0
     end
 
     # Replies to this message with the specified content.
@@ -290,6 +294,15 @@ module Discordrb
     # @return [Array<Reaction>] the reactions
     def my_reactions
       @reactions.select(&:me)
+    end
+
+    # Removes embeds from the message
+    # @return [Message] the resulting message.
+    def suppress_embeds
+      Message.new(
+        JSON.parse(API::Channel.suppress_embeds(@bot.token, @channel.id, @id)),
+        @bot
+      )
     end
 
     # Reacts to a message.
--- a/lib/discordrb/api/channel.rb
+++ b/lib/discordrb/api/channel.rb
@@ -129,6 +129,19 @@ module Discordrb::API::Channel
     )
   end
 
+  # Suppress embeds on a message
+  def suppress_embeds(token, channel_id, message_id)
+    Discordrb::API.request(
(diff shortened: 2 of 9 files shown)


lib/discordrb/data/message.rb
```diff
@@ -286,6 +290,13 @@ def my_reactions
       @reactions.select(&:me)
     end
 
+    # Removes embeds from the message
+    # @return [Message] the resulting message.
+    def suppress_embeds
+      response = API::Channel.suppress_embeds(@bot.token, @channel.id, @id)
```
Droid00000: I believe it might be better to wrap the entire API call in in ``JSON.parse``. i.e. ``JSON.parse(API::Channel.suppress_embeds(@bot.token, @channel.id, @id))

discordrb.gemspec
```diff
@@ -44,4 +44,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rubocop-rake', '~> 0.6.0'
   spec.add_development_dependency 'simplecov', '~> 0.21.0'
   spec.add_development_dependency 'yard', '~> 0.9.9'
+  spec.add_development_dependency 'pry-byebug'
```
Droid00000: I don't think this dependency is really needed, but, if there is a valid use-case, feel-free to start a discussion in the [drb channel.](https://discord.gg/discord-api)

Droid00000 (changes requested):
Mainly looks good just needs a few minor changes. The main thing is that I'd say the `pry-byebug` dependency isn't really needed and goes way beyond the scope of this PR.

Quintasan:
> Mainly looks good just needs a few minor changes. The main thing is that I'd say the `pry-byebug` dependency isn't really needed and goes way beyond the scope of this PR.

We should be good to go now

Droid00000:
> > Mainly looks good just needs a few minor changes. The main thing is that I'd say the `pry-byebug` dependency isn't really needed and goes way beyond the scope of this PR.
> 
> 
> 
> We should be good to go now

Almost there, I think you forgot to remove the dependency from the spec dependency though 😅

Droid00000 (approved):

Quintasan:
@Droid00000 well, this should be good now

wouterdedroog:
Looks great! It could be interesting to use this PR to further implement message flags (in a similar way to how the permissions class is currently structured). This is what I came up with: https://gist.github.com/wouterdedroog/59ed3076558c6d943dca5f8c75d5437f

Quintasan:
@Droid00000 bump

swarley:
Would you be open to doing the API portion through a message update rather than a dedicated API method? I don't want to add an API level method for every flag that may need to be updated. Just something like
```ruby
flags = @flags | SUPPRESS_EMBEDS
API::Channel.edit_message(..., flags)
```

Quintasan:
> Would you be open to doing the API portion through a message update rather than a dedicated API method? I don't want to add an API level method for every flag that may need to be updated. Just something like
> 
> ```ruby
> flags = @flags | SUPPRESS_EMBEDS
> API::Channel.edit_message(..., flags)
> ```

I most certainly don't mind, but I'd like to work on this incrementally. When this gets merged, then I'll prepare another PR

PR #242, opened 2023-10-18
feat: chose a specific type in channel_selection

itakerboust:
# Summary

This feature allows filtering the possible channel types when using the "channel_select" method.

| Before (display all) | After (display text & voice channel) |
|----------|----------|
| image | image |

## Changed

* Add new parameter _channel_types_ in `channel_select` method
* Include _channel_types_ parameter in example


lib/discordrb/webhooks/view.rb
```diff
@@ -24,6 +24,24 @@ class Discordrb::Webhooks::View
...
+    guild_category: 4,
+    guild_announcement: 5,
+    announcement_thread: 10,
+    public_thread: 11,
+    private_thread: 12,
+    guild_stage_voice: 13,
+    guild_directory: 14,
+    guild_forum: 15,
+    guild_media: 16
+  }.freeze
```
Dakurei: Perhaps it would be better to reuse the existing constant and update it, rather than having a duplicate with different values ?
(By "update it", I mean add what's obviously missing. We'll avoid provoking a big breaking change by modifying existing content though 😅)

https://github.com/shardlab/discordrb/blob/57c1f5c665df12ad537f7e42cd6213af892fa1f4/lib/discordrb/data/channel.rb#L11-L26

idrista: Thank you, I hadn't found the existing one

Dakurei (approved):

Droid00000:
Sorry we missed this, but this was implemented in #297. Thank you for the effort though!

PR #249, opened 2024-01-08
Fix buttons and string selects without emojis breaking due to API change

### LINKED ISSUE #247: Buttons (and string selects) without emojis broken due to a Discord API change (easy fix)
### Description:
Starting today, buttons and string select options without explicit emojis are failing, returning the following error:
```
Invalid Form Body
    - components[0].components[0].emoji.name: Invalid emoji
Backtrace:
/.../lib/discordrb/api.rb:124:in `rescue in request'
/.../lib/discordrb/api.rb:113:in `request'
/.../lib/discordrb/api/channel.rb:92:in `create_message'
/.../lib/discordrb/bot.rb:405:in `send_message'
/.../lib/discordrb/data/channel.rb:433:in `send_message'
...
```
The issue appears to be that `{}` is no longer accepted as a default value for the emoji field, instead requiring an explicit `null` value.
### Solution:
An easy tested fix is just to add the extra case
```ruby
when nil
  nil
```
in the button builder: https://github.com/shardlab/discordrb/blob/0f01116c3326e4784a0b581326307f5a2b57427a/lib/discordrb/webhooks/view.rb#L48-L53
The same fix works for string select options here: https://github.com/shardlab/discordrb/blob/0f01116c3326e4784a0b581326307f5a2b57427a/lib/discordrb/webhooks/view.rb#L148-L153

MrWouterNL:
Fix buttons and string selects without emojis breaking due to API change (fixes #247)

# Summary

Due to a Discord API change, buttons and string selects without emojis no longer work in the current Discordrb version (as stated in #247. This PR fixes this issue by returning nil instead of an empty hash.

### DIFF
--- a/lib/discordrb/webhooks/view.rb
+++ b/lib/discordrb/webhooks/view.rb
@@ -49,7 +49,7 @@ class Discordrb::Webhooks::View
               when Integer, String
                 emoji.to_i.positive? ? { id: emoji } : { name: emoji }
               else
-                emoji.to_h
+                emoji&.to_h
               end
 
       @components << { type: COMPONENT_TYPES[:button], label: label, emoji: emoji, style: style, custom_id: custom_id, disabled: disabled, url: url }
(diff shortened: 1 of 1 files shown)


swarley:
Thanks!

PR #251, opened 2024-01-23
Fix InteractionCreateEvent type matching

viacommits:
# Summary

This PR fixes some bugs where matching Interactions based on type via a symbol or string wasn't working. There was a typo in the matcher block, and as far as I can call the comparison wasn't working either. Specs have been added to cover all formats described as supported in the documentation for matching Interaction types (Integer, String, and Symbol).

Example broken code before this fix:

```rb
module EventHandler
  extend Discordrb::EventContainer

  interaction_create type: :component do |event|
    event.defer
  end
end
```

This would result in `NameError: uninitialized constant Discordrb::Interactions::TYPES`

---

## Added

* Specs for `InteractionCreateEvent`

## Changed

* Matching behavior for `:type` attribute in `InteractionCreateEventHandler`

## Fixed

* `NameError: uninitialized constant Discordrb::Interactions::TYPES` in `InteractionCreateEventHandler` when matching on `:type`
* Incorrect matching logic with Strings and Symbols in `InteractionCreateEventHandler` `:type` attribute

### DIFF
--- a/lib/discordrb/events/interactions.rb
+++ b/lib/discordrb/events/interactions.rb
@@ -105,11 +105,11 @@ module Discordrb::Events
       [
         matches_all(@attributes[:type], event.type) do |a, e|
           match_type = case a
-               when String, Symbol
-                 Discordrb::Interaction::TYPES[a.to_sym]
-               else
-                 a
-               end
+                       when String, Symbol
+                         Discordrb::Interaction::TYPES[a.to_sym]
+                       else
+                         a
+                       end
           match_type == e
         end,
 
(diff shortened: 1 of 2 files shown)


zeiv:
This should be ready to review, it appears that the Ruby 3.1 failures are unrelated to this PR and due to libsodium issues. Rubocop violations have been addressed though I can't rerun the check.

lib/discordrb/events/interactions.rb
```diff
@@ -104,12 +104,13 @@ def matches?(event)
 
       [
         matches_all(@attributes[:type], event.type) do |a, e|
-          a == case a
-               when String, Symbol
-                 Discordrb::Interactions::TYPES[e.to_sym]
-               else
-                 e
-               end
+          match_type = case a
```
Droid00000, line 107: I feel like we could just do:

```ruby
case a
when String, Symbol
  Discordrb::Interactions::TYPES[a.to_sym] == e
else
  a == e
end
```

PR #255, opened 2024-03-18
Fix edit_application_command for global commands

MrWouterNL:


MrWouterNL:
Oops, this is a duplicate.

PR #262, opened 2024-06-04
Add support for role mentions on the bot's autogenerated managed roleid.

Ifiht:
# Summary

This is to address the discrepancy when a user types the bot's exact name, and discord translates it to a role mention due to the autogenerated managed role that's created when adding a bot with permissions > 0 to a server. Below is an example of the new behavior from an `irb` session:
```Ruby
3.1.4 :003 > bot.run :async
[INFO : websocket @ 2024-06-05 09:54:55.116] Discord using gateway protocol version: 9, requested: 9
 => nil
3.1.4 :004 > bot.mention(allow_role_mention: true) do |event|
3.1.4 :005 >     puts "THIS FIRES FOR USER & ROLE MENTIONS: #{event.content}"
3.1.4 :006 > end
 => #<Discordrb::Events::MentionEventHandler:0x000055742aa345c8 @attributes={:allow_role_mention=>true}, @block=#<Proc:0x000055742aa34a00 (irb):4>>
3.1.4 :007 > bot.mention() do |event|
3.1.4 :008 >     puts "THIS ONLY FIRES FOR USER MENTIONS: #{event.content}"
3.1.4 :009 > end
 => #<Discordrb::Events::MentionEventHandler:0x000055742ab280d8 @attributes={}, @block=#<Proc:0x000055742ab282b8 (irb):7>>
3.1.4 :010 >
3.1.4 :011 > 
THIS FIRES FOR USER & ROLE MENTIONS: <@1211423563475849236> user mention test
THIS ONLY FIRES FOR USER MENTIONS: <@1211423563475849236> user mention test
THIS FIRES FOR USER & ROLE MENTIONS: <@&1211432785353637999> role mention test
```

in the above test, I sent two messages to the bot, one with a role mention and the other with a user mention. The blocks passed demonstrate that _only_ when `:allow_role_mention == true` will the event be raised upon role mentions.

All tests were run with `bundle exec rspec spec` and passed:
```
Finished in 0.27027 seconds (files took 0.38679 seconds to load)
430 examples, 0 failures, 3 pending
```

---

## Changed
-  `discordrb/events/message.rb`
- `discordrb/bot.rb`
- `discordrb/container.rb`

additionally the gemspec was updated to include the `role_mentions:` flag, to avoid a no method error but otherwise function is exactly the same.

## Fixed

You could say the `bot.mention` function has been "fixed"? Depends how you view this feature & support for Discord's managed roles.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -1321,7 +1321,7 @@ module Discordrb
         end
         # Raises a MentionEvent for a bot's managed integration role:
         # A managed role is auto-generated for any bots joined to a server with permissions > 0
-        # For this to generate an event, 
+        # For this to generate an event, set role_managed: true
         if message.role_mentions.any? { |role| role.managed? == true && role.tags.bot_id == @profile.id }
           event = MentionEvent.new(message, self)
           event.role_mention = true
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -1317,6 +1317,15 @@ module Discordrb
 
         if message.mentions.any? { |user| user.id == @profile.id }
           event = MentionEvent.new(message, self)
+          event.role_mention = false
+          raise_event(event)
+        end
+        # Raises a MentionEvent for a bot's managed integration role:
+        # A managed role is auto-generated for any bots joined to a server with permissions > 0
+        # For this to generate an event, 
+        if message.role_mentions.any? { |role| role.managed? == true && role.tags.bot_id == @profile.id }
+          event = MentionEvent.new(message, self)
+          event.role_mention = true
           raise_event(event)
         end
 
--- a/lib/discordrb/events/message.rb
+++ b/lib/discordrb/events/message.rb
@@ -264,10 +264,92 @@ module Discordrb::Events
   end
 
   # @see Discordrb::EventContainer#mention
-  class MentionEvent < MessageEvent; end
+  class MentionEvent < MessageEvent
+    attr_accessor :role_mention
+
+    def initialize(message, bot)
+      @bot = bot
+      @message = message
+      @channel = message.channel
+      @saved_message = ''
+      @file = nil
+      @filename = nil
+      @file_spoiler = nil
+      @role_mention = false
+    end
+  end
 
   # Event handler for {MentionEvent}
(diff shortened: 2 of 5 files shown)


lib/discordrb/bot.rb
```diff
@@ -1316,6 +1316,15 @@ def handle_dispatch(type, data)
...
           event = MentionEvent.new(message, self)
+          event.role_mention = false
+          raise_event(event)
+        end
+        # Raises a MentionEvent for a bot's managed integration role:
+        # A managed role is auto-generated for any bots joined to a server with permissions > 0
+        # For this to generate an event, set role_managed: true
+        if message.role_mentions.any? { |role| role.managed? == true && role.tags.bot_id == @profile.id }
+          event = MentionEvent.new(message, self)
+          event.role_mention = true
```
Daniel-Worrall, line 1327: Not a fan of this pattern and we don't do it anywhere else. Let's put this in the MentionEvent constructor and do `MentionEvent.new(message, self, false)`

lib/discordrb/events/message.rb
```diff
@@ -264,10 +264,92 @@ def after_call(event)
   end
 
   # @see Discordrb::EventContainer#mention
-  class MentionEvent < MessageEvent; end
+  class MentionEvent < MessageEvent
+    attr_accessor :role_mention
+
+    def initialize(message, bot)
```
Daniel-Worrall, line 270: We should call super here instead of setting all these again

lib/discordrb/version.rb
```diff
@@ -3,5 +3,5 @@
 # Discordrb and all its functionality, in this case only the version.
 module Discordrb
   # The current version of discordrb.
-  VERSION = '3.5.0'
+  VERSION = '3.5.1'
```
Daniel-Worrall, line 6: Please don't include version changes in PRs

Daniel-Worrall (changes requested):
Approved other than comments

Daniel-Worrall:
@Ifiht That's perfectly understandable - If it's okay with you, I'd like to commit these changes on your behalf with these changes

Ifiht:
Sure please use however you'd like! Sorry I just don't have time to go back and refresh this, already got rid of everything I was working on here.

Issue #256, opened 2024-04-18
Ruby 3.3 deprecation warnings: base64 will no longer be part of the default gems


y-bonfire:
# Summary

```
ruby/3.3.0/gems/discordrb-3.5.0/lib/discordrb/data.rb:18: warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec. Also contact author of discordrb-3.5.0 to add base64 into its gemspec.
```

I reported it as written like this.
>  Also contact author of discordrb-3.5.0 to add base64 into its gemspec.

I confirmed that there are no identical issues or pull requests through searching.

https://github.com/ruby/ruby/commit/6500f85927135901f365d2ba1186c5c4f2881f8e

I thought that adding this line would probably be the solution.
https://github.com/shardlab/discordrb/blob/main/discordrb.gemspec
```ruby
 spec.add_runtime_dependency 'base64'
```

---

## Environment

**Ruby version:**

```
ruby -v
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-linux]
```

**Discordrb version:**

```
$ bundle list |grep discord
  * discordrb (3.5.0)
```

PixeLInc:
fixed in merge of: https://github.com/shardlab/discordrb/pull/277

PR #263, opened 2024-06-08
Add base64 gem explicitly as dependency

mataku:
# Summary

closes https://github.com/shardlab/discordrb/issues/256

base64 is no longer the default gem from Ruby 3.4
https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/

On Ruby 3.3, we could see the warning:

```ruby
warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec.
```

From Ruby 3.4, we will get the error:

```ruby
Failure/Error: require 'base64'

LoadError:
  cannot load such file -- base64
```

---

## Added

Add base64 dependency to gemspec

### DIFF
--- a/discordrb.gemspec
+++ b/discordrb.gemspec
@@ -34,6 +34,8 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'discordrb-webhooks', '~> 3.5.0'
 
+  spec.add_dependency 'base64', '~> 0.2.0'
+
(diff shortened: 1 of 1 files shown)


PixeLInc:
This fix was included in another PR that was merged: https://github.com/shardlab/discordrb/pull/277

so going to close this one out. Thanks for the PR, though.

PR #264, opened 2024-06-09
Update webhook URLs to v9

advaith1:
# Summary

discordrb uses API v9 but the generated webhook URLs use API v8. This PR updates webhook URLs to v9 for consistency. Webhooks don't get breaking changes anyway

## Fixed

updated webhook URLs from v8 to v9


swarley (approved):

swarley:
Thanks!

Issue #238, opened 2023-09-26
`Channel#permission_overwrites=` argument type mismatched with docs


anna328p:
# Summary

Here's the documentation for `Channel#permission_overwrites=` (as rendered by YARD):

> **`#permission_overwrites=`**`(overwrites) ⇒ Object`
> ﻿
> Bulk sets this channels permission overwrites
> ﻿
> **Parameters:**
> - **`overwrites`** (`Array<Overwrite>`)

When I try to call the method as documented (`channel.permission_overwrites = [ow1, ow2]`), I get a stack trace like the following:

```
Exception: #<NoMethodError: undefined method `to_hash' for nil:NilClass>
[...]/lib/discordrb/data/channel.rb:968:in `block in update_channel_data'
[...]/lib/discordrb/data/channel.rb:968:in `map'
[...]/lib/discordrb/data/channel.rb:968:in `update_channel_data'
[...]/lib/discordrb/data/channel.rb:346:in `permission_overwrites='
(eval):1:in `block in <module:Admin>'
[...]/qbot/modules/admin.rb:45:in `eval'
[...]/qbot/modules/admin.rb:45:in `block in <module:Admin>'
[...]/lib/discordrb/commands/parser.rb:108:in `call'
[...]/qbot/lib/qbot/patches.rb:35:in `call'
[...]/lib/discordrb/commands/command_bot.rb:240:in `execute_command'
[...]/qbot/lib/qbot/hooks.rb:30:in `execute_command'
[...]/lib/discordrb/commands/command_bot.rb:340:in `simple_execute'
[...]/lib/discordrb/commands/command_bot.rb:498:in `block in execute_chain'
```

This is because the implementation of this method doesn't match the docs. Here's the code for that method:

https://github.com/shardlab/discordrb/blob/a873d4efad8e87af5732a7e63044ff95cebc11d6/lib/discordrb/data/channel.rb#L345-L347

Here's the code for the method it calls:

https://github.com/shardlab/discordrb/blob/a873d4efad8e87af5732a7e63044ff95cebc11d6/lib/discordrb/data/channel.rb#L965-L980

The expression `new_data[:permission_overwrites].map { |_, v| v.to_hash }` is present here, which I infer assumes `new_data[:permission_overwrites]` is a `Hash` (or something else that can be destructured into pairs). However, this makes no sense semantically, since the relevant Discord API endpoint (PATCH `/channels/{channel.id}`) takes an _array_ of permission overwrite objects.

---

## Environment

[...]

PR #267, opened 2024-07-10
Fix permission overwrites mapping

ceud:
Fix for https://github.com/shardlab/discordrb/issues/238


Daniel-Worrall (approved):
Thank you, seems this has been broken for a while!

PR #279, opened 2024-10-16
Fix Guild Bans

Droid00000:
# Summary
Updates the ban API call, abstractions, and inline docs, to support cleaning messages as the old `delete_message_days` parameter has been replaced by `delete_message_seconds`, and adds support for sending an X-Audit-Log-Reason header.
## Fixed
No audit log reason for guild bans, and broken clean messages parameter.
## Added
Adds support for bulk banning.


lib/discordrb/data/member.rb
```diff
@@ -247,10 +247,10 @@ def server_unmute
     end
 
     # Bans this member from the server.
-    # @param message_days [Integer] How many days worth of messages sent by the member should be deleted.
+    # @param message_seconds [Integer] How many seconds worth of messages sent by the member should be deleted.
     # @param reason [String] The reason this member is being banned.
-    def ban(message_days = 0, reason: nil)
-      server.ban(@user, message_days, reason: reason)
+    def ban(message_seconds = 0, reason: nil)
```
expeehaa: I would prefer if this breaking change happened in version 4.0, since it's such an easy thing to overlook if one doesn't expect a breaking change. For 3.x, I guess it would be better to support floating point numbers for `message_days`, multiply them with 86400 and turn that into an integer to pass on to the API.

lib/discordrb/api/server.rb
```diff
@@ -200,17 +200,18 @@ def bans(token, server_id, limit = nil, before = nil, after = nil)
     )
   end
 
-  # Ban a user from a server and delete their messages from the last message_days days
+  # Ban a user from a server and delete their messages from seconds.
```
expeehaa: This change does not make sense to me. Did you mean to write ".. the last message_seconds seconds."?

lib/discordrb/data/server.rb
```diff
@@ -610,10 +610,10 @@ def bans(limit: nil, before_id: nil, after_id: nil)
 
     # Bans a user from this server.
     # @param user [User, String, Integer] The user to ban.
-    # @param message_days [Integer] How many days worth of messages sent by the user should be deleted.
+    # @param message_seconds [Integer] How many seconds of messages sent by the user should be deleted.
     # @param reason [String] The reason the user is being banned.
-    def ban(user, message_days = 0, reason: nil)
-      API::Server.ban_user(@bot.token, @id, user.resolve_id, message_days, reason)
+    def ban(user, message_seconds = 0, reason: nil)
```
expeehaa: Same as above.

lib/discordrb/api/server.rb
Droid00000: I believe that originally "the last message_days days" would've been incorrect. I don't believe there's any need to list the parameter name in the description as it doesn't seem to be standard practice for this branch. It also doesn't make much sense, grammatically.

Droid00000:
I've added back support for the `message_days` parameter. I've included `message_seconds` as an optional parameter to ease the transition to 4.0, when that eventually releases.

lib/discordrb/data/member.rb
```diff
@@ -247,10 +247,17 @@ def server_unmute
     end
 
     # Bans this member from the server.
-    # @param message_days [Integer] How many days worth of messages sent by the member should be deleted.
+    # @param message_days [Integer] How many days worth of messages sent by the member should be deleted. This will be deprecated in 4.0.
+    # @param message_seconds [Integer] How many seconds worth of messages sent by the member should be deleted.
     # @param reason [String] The reason this member is being banned.
-    def ban(message_days = 0, reason: nil)
-      server.ban(@user, message_days, reason: reason)
+    def ban(message_days = 0, message_seconds: nil, reason: nil)
```
expeehaa, line 309: I don’t like `message_seconds` as a keyword argument, but I suppose it’s the only reasonable way to achieve being able to choose between units.

lib/discordrb/data/member.rb
```diff
@@ -247,10 +247,17 @@ def server_unmute
     end
 
     # Bans this member from the server.
-    # @param message_days [Integer] How many days worth of messages sent by the member should be deleted.
+    # @param message_days [Integer] How many days worth of messages sent by the member should be deleted. This will be deprecated in 4.0.
```
expeehaa: Shouldn’t this parameter be deprecated now and removed in 4.0?

lib/discordrb/api/server.rb
expeehaa: I agree that the previous comment could have been worded better, but "delete their messages from seconds" sounds incorrect, since "seconds" isn’t some object that contains "their messages", such that they could be deleted from it. Rather, the user’s messages written after a point in time that was the specified amount of seconds ago are deleted.
How’ld something like `... and delete their messages written up to a given amount of time ago.` be?

lib/discordrb/data/server.rb
```diff
@@ -610,10 +610,17 @@ def bans(limit: nil, before_id: nil, after_id: nil)
 
     # Bans a user from this server.
     # @param user [User, String, Integer] The user to ban.
-    # @param message_days [Integer] How many days worth of messages sent by the user should be deleted.
+    # @param message_days [Integer] How many days worth of messages sent by the user should be deleted. This will be deprecated in 4.0.
```
expeehaa: Same as above.

expeehaa (changes requested):

lib/discordrb/api/server.rb
Droid00000: I have updated the wording to more closely align with what you've suggested.

lib/discordrb/data/member.rb
Droid00000: As per what @swarley has [said in DAPI](https://discord.com/channels/81384788765712384/381891448884428801/1297344043012980807) I was told to keep the argument as is and add seconds as an optional KWARG.

Droid00000 (commented):
I've made a few changes and responded to most of the comments. Please take a look whenever you have the chance. 🙇‍♂️

lib/discordrb/data/member.rb
expeehaa: I think that comment referred to keeping it in version `3.x`, not in `4.0`. In semantic versioning (semver), a breaking change, like a backwards incompatible method signature change, is only allowed during major version changes, such as `3.x` to `4.0`. A deprecation is a warning that it will be removed in a future version, in this case likely `4.0`. Therefore, it should be deprecated now and removed in `4.0`, not, as the comment claims, be deprecated in `4.0`, which would imply its existence in `4.0` (and thus its removal not earlier than in `5.0`).

Droid00000: > I think that comment referred to keeping it in version `3.x`, not in `4.0`. In semantic versioning (semver), a breaking change, like a backwards incompatible method signature change, is only allowed during major version changes, such as `3.x` to `4.0`. A deprecation is a warning that it will be removed in a future version, in this case likely `4.0`. Therefore, it should be deprecated now and removed in `4.0`, not, as the comment claims, be deprecated in `4.0`, which would imply its existence in `4.0` (and thus its removal not earlier than in `5.0`).

Oh whoops. Seems like I misunderstood what you mean.

expeehaa (approved):

lib/discordrb/api/server.rb
```diff
@@ -200,17 +200,18 @@ def bans(token, server_id, limit = nil, before = nil, after = nil)
...
-      "#{Discordrb::API.api_base}/guilds/#{server_id}/bans/#{user_id}?delete_message_days=#{message_days}&reason=#{reason}",
-      nil,
-      Authorization: token
+      "#{Discordrb::API.api_base}/guilds/#{server_id}/bans/#{user_id}",
+      { delete_message_seconds: message_seconds }.to_json,
+      Authorization: token,
+      content_type: :json,
+      'X-Audit-Log-Reason': reason
     )
   end
```
swarley: Unfortunately API scoped methods are techincally public interfaces. Let's make prime versions of these that work with seconds, and make the existing ones convert days to seconds

i.e.

```ruby
def ban_user(token, server_id, user_id, message_days, reason = nil)
    ban_user!(token, server_id, user_id, message_days * (60 * 60 * 24), reason)
end

def ban_user!(token, server_id, user_id, message_seconds, reason = nil)
   #....
 end
 ```

swarley (changes requested):

lib/discordrb/api/server.rb
Droid00000: alr, went and made a prime variant of this

swarley (approved):

PR #282, opened 2024-10-25
feat: application emojis

Droid00000:
# Summary
Add support for CRUD operations on application emojis via the REST API.

## Added
API calls for application emojis.
`API::Application.list_application_emojis`
`API::Application.get_application_emoji`
`API::Application.create_application_emoji`
`API::Application.edit_application_emoji`
`API::Application.delete_application_emoji`

Abstractions for application emojis.
`Bot#application_emojis`
`Bot#get_application_emoji`
`Bot#create_application_emoji`
`Bot#edit_application_emoji`
`Bot#delete_application_emoji`

### DIFF
--- a/lib/discordrb/api/application.rb
+++ b/lib/discordrb/api/application.rb
@@ -199,4 +199,68 @@ module Discordrb::API::Application
       content_type: :json
     )
   end
+
+  # Get a list of application emojis.
+  # https://discord.com/developers/docs/resources/emoji#list-application-emojis
+  def list_application_emojis(token, application_id)
+    Discordrb::API.request(
+      :applications_aid_emojis,
+      nil,
+      :get,
+      "#{Discordrb::API.api_base}/applications/#{application_id}/emojis",
+      Authorization: token
+    )
+  end
+
+  # Get an application emoji by ID.
+  # https://discord.com/developers/docs/resources/emoji#get-application-emoji
+  def get_application_emoji(token, application_id, emoji_id)
+    Discordrb::API.request(
+      :applications_aid_emojis_eid,
+      nil,
+      :get,
+      "#{Discordrb::API.api_base}/applications/#{application_id}/emojis/#{emoji_id}",
+      Authorization: token
+    )
+  end
+
+  # Create an application emoji.
+  # https://discord.com/developers/docs/resources/emoji#create-application-emoji
+  def create_application_emoji(token, application_id, name, image)
+    Discordrb::API.request(
+      :applications_aid_emojis,
+      nil,
+      :post,
+      "#{Discordrb::API.api_base}/applications/#{application_id}/emojis",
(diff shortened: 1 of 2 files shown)


vStone:
Should it not be `image/jpeg` as per the docs? https://discord.com/developers/docs/reference#image-data

Also, maybe support for gif and png image uploads (by file name extension) could be useful

swarley:
Looks like this logic was pulled from the standard emoji flow which does this poorly. We do already have a MIME type gem which is used for role icons

```
      mime_type = MIME::Types.type_for(icon.__send__(path_method)).first&.to_s || 'image/jpeg'
      ```

**Comment by Droid00000** [comment]
I will update the emoji flow to use MIME types.

**Comment by Droid00000** [comment]
> Looks like this logic was pulled from the standard emoji flow which does this poorly. We do already have a MIME type gem which is used for role icons
> 
> ```ruby
>       mime_type = MIME::Types.type_for(icon.__send__(path_method)).first&.to_s || 'image/jpeg'
> ```

I have pushed a commit to make the `Bot#create_application_emoji` method utilize MIME types.

**Comment by vStone** [comment]
For documentation purposes, if you need to resize before uploading:

```ruby
ImageWrapper = Struct.new(:blob, :original_filename_value) do
  def read
    blob
  end

  def original_filename
    original_filename_value
  end
end
```

```ruby
image = MiniMagick::Image.open(file_path)
image.resize('256x256')
image_wrapper = ImageWrapper.new(image.to_blob, file_path)
emoji = @bot.create_application_emoji(emoji_name, image_wrapper)
```

Droid00000:
> For documentation purposes, if you need to resize before uploading:
> 
> 
> 
> ```ruby
> 
> ImageWrapper = Struct.new(:blob, :original_filename_value) do
> 
>   def read
> 
>     blob
> 
>   end
> 
> 
> 
>   def original_filename
> 
>     original_filename_value
> 
>   end
> 
> end
> 
> ```
> 
> 
> 
> ```ruby
> 
> image = MiniMagick::Image.open(file_path)
> 
> image.resize('256x256')
> 
> image_wrapper = ImageWrapper.new(image.to_blob, file_path)
> 
> emoji = @bot.create_application_emoji(emoji_name, image_wrapper)
> 
> ```
> 
> 

Is this a request for an addition to the docs or is this comment meant to serve as the documentation itself.

vStone:
I needed to brain-dump this somewhere for future-self. You could add it to the docs or throw it in an example perhaps?

Droid00000:
> I needed to brain-dump this somewhere for future-self. You could add it to the docs or throw it in an example perhaps?
> 
> 
> 
> 

Ah, alright, that's cool by me. I won't be adding this to the docs or making an example out of it, though. It would be better to just point people to your comment, imo.

Issue #60, opened 2021-06-01
Channel#reacted_with limit:nil Bug - Again


Daniel-Worrall:
# Summary

We fixed the Pagination without supplying a limit in #39, but discord has recently broke this.

https://github.com/discord/discord-api-docs/issues/3053

An untested fix in the meantime could be to supply a known limit, but it'll still make additional requests and you'd have to call .uniq on the Array.

You could also supply custom pagination code.

y-bonfire:
Only retrieve 100 results, regardless of whether 'nil' or a specific number like 200 was specified.  
After rewriting it using another language, another library, the issue was resolved, so I believe this is a bug.

Creating a message with more than 100 reactions is difficult, so I think it's challenging to test this behavior.

PR #288, opened 2024-11-29
Bugfix - Discordrb::Message#reacted_with

y-bonfire:
# Summary

I noticed this bug recently.
https://github.com/shardlab/discordrb/issues/60

Upon investigation, I realized that the issue stems from a misunderstanding of how `proc` works. Returning inside a `proc` causes it to return a value immediately, which prevents the loop from continuing. This appears to be a basic Ruby mistake.

After applying the fix, I ran the code locally and confirmed that it successfully retrieves users with more than 100 reactions.

Here is a simple sample to illustrate the issue.
```ruby

def proc_return_test

  proc_test = proc do |a|
    return a
  end

  val = proc_test.call(1)
  puts val
  0
end

def proc_next_test

  proc_test = proc do |a|
    next a
  end

  val = proc_test.call(1)
  puts val
  0
end


# 1
puts proc_return_test

# 1,0
puts proc_next_test


```

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -314,7 +314,7 @@ module Discordrb
 
       get_reactions = proc do |fetch_limit, after_id = nil|
         resp = API::Channel.get_reactions(@bot.token, @channel.id, @id, reaction, nil, after_id, fetch_limit)
-        return JSON.parse(resp).map { |d| User.new(d, @bot) }
+        JSON.parse(resp).map { |d| User.new(d, @bot) }
       end
 
       # Can be done without pagination


y-bonfire:
@swarley @Daniel-Worrall  review please

swarley (approved):

Daniel-Worrall:
@y-bonfire Was the code tested before the change as well? The code changed in the PR originates after the issue was posted, and the issue may not even still be applicable

y-bonfire:
> @y-bonfire Was the code tested before the change as well? The code changed in the PR originates after the issue was posted, and the issue may not even still be applicable

@Daniel-Worrall 
Yes, I confirmed that for cases like 600 reactions, the code previously only retrieved 100. After the change, it now retrieves the expected values as intended.

Daniel-Worrall (approved):

y-bonfire:
@Daniel-Worrall 
Thanks a lot for merging this!  
Really appreciate your time and effort.😊

PR #294, opened 2024-12-22
feature: channel pins update event

Droid00000:
# Summary

Adds support for receiving and listening for the channel pins update gateway event.

## Added
`Discordrb::EventContainer#channel_pins_update`
`ChannelPinsUpdateEvent`
`ChannelPinsUpdateEventHandler`


swarley:
I was going to ask for nice to haves like splitting this into `pinned`/`unpinned` or some filtering attributes but after reviewing the docs it looks like this is just another certified discord banger that doesn't include that information

swarley:
Thanks!

PR #297, opened 2025-02-24
feat: new components

Droid00000:
## Summary

This is my attempt at adding support for the new components that Discord plans to release in the distant future. This introduces the ability to have components besides action rows as top-level components, as well as the ability to nest action rows within other top-level components. Some potential points to discuss:

- The example admittedly isn't the best. I'm open to doing something different than what I've current added.
- `new_components` admittedly isn't the best name for setting the `IS_COMPONENTS_V2 (1 << 15)` flag.

## Added

- Builders for new components:
`Webhooks::View::TextDisplayBuilder`
`Webhooks::View::SeperatorBuilder`
`Webhooks::View::FileBuilder`
`Webhooks::View::MediaGalleryBuilder`
`Webhooks::View::SectionBuilder`
`Webhooks::View::ContainerBuilder`

- Top-level methods for new components:
`Webhooks::View#text_display`
`Webhooks::View#seperator`
`Webhooks::View#file`
`Webhooks::View#media_gallery`
`Webhooks::View#section`
`Webhooks::View#container`

- Data classes for serializing new components:
`Webhooks::View::Components::TextDisplay`
`Webhooks::View::Components::Seperator`
`Webhooks::View::Components::File`
`Webhooks::View::Components::MediaGallery`
`Webhooks::View::Components::Section`
`Webhooks::View::Components::Container`

`new_components` KWARG for interaction methods which sets the `IS_COMPONENTS_V2 (1 << 15)` flag.

## Changed

There are a couple of internal changes made to `Webhooks::View` in order to make the new components possible. (Don't worry, I've managed to maintain full backwards compatibility!)

Removal of `@rows`. Since action rows are no longer the only top-level components, it didn't make sense to have two separate instance variables since it would end up requiring some needless concatenation. E.g.

```ruby
[@rows.map(&:to_h), *@components.map(&:to_h)].reject(&:empty)
```

In its place, I've replaced it with `@components` as a way to store all top-level components. This required the removal of the `attr_reader :rows` as well. It was replaced with:

```ruby
def rows
  @components.select { |component| component.is_a?(RowBuilder) }
end
```

## Fixed

Previously, when passing in an emoji object to `RowBuilder#button` and `SelectMenuBuilder#option`, it would yield the following error:

```ruby
Exception: #<NoMethodError: undefined method `to_h' for an instance of Discordrb::Emoji>
```

As stated in the error, this is because `#to_h` wasn't defined anywhere for the emoji object. I've went ahead and defined this method.

### DIFF
--- a/lib/discordrb/data/component.rb
+++ b/lib/discordrb/data/component.rb
@@ -18,6 +18,20 @@ module Discordrb
         SelectMenu.new(data, bot)
       when Webhooks::Modal::COMPONENT_TYPES[:text_input]
         TextInput.new(data, bot)
+      when Webhooks::View::COMPONENT_TYPES[:section]
+        Section.new(data, bot)
+      when Webhooks::View::COMPONENT_TYPES[:text_display]
+        TextDisplay.new(data, bot)
+      when Webhooks::View::COMPONENT_TYPES[:thumbnail]
+        Thumbnail.new(data, bot)
+      when Webhooks::View::COMPONENT_TYPES[:media_gallery]
+        MediaGallery.new(data, bot)
+      when Webhooks::View::COMPONENT_TYPES[:file]
+        File.new(data, bot)
+      when Webhooks::View::COMPONENT_TYPES[:seperator]
+        Seperator.new(data, bot)
+      when Webhooks::View::COMPONENT_TYPES[:container]
+        Container.new(data, bot)
       end
     end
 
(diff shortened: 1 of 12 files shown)


lib/discordrb/webhooks/modal.rb
```diff
@@ -5,76 +5,197 @@ class Discordrb::Webhooks::Modal
...
+    file_upload: 19
   }.freeze
 
-  # This builder is used when constructing an ActionRow. All current components must be within an action row, but this can
-  # change in the future. A message can have 5 action rows, each action row can hold a weight of 5. Buttons have a weight of 1,
-  # and dropdowns have a weight of 5.
-  class RowBuilder
-    # A mapping of short names to types of input styles. `short` is a single line where `paragraph` is a block.
+  # This builder is used when adding a label component to a modal.
+  class LabelBuilder
```
swarley, line 74: i wonder if it's worth it for semver purposes so make labelbuilder extend row builder because it would break is_a? checks.. but i also don't know why someone would be doing that necessarily. we should probably strive for keeping our semver contract regardless

lib/discordrb/data/component.rb
```diff
@@ -14,23 +14,45 @@ def self.from_data(data, bot)
         ActionRow.new(data, bot)
       when Webhooks::View::COMPONENT_TYPES[:button]
         Button.new(data, bot)
-      when Webhooks::View::COMPONENT_TYPES[:string_select]
+      when 3, 5, 6, 7, 8 # Select Menu Types
```
swarley: i assume this is to avoid a big line but i would still like to see a better match here to keep it consistent

Droid00000: Done. This uses named constants now

lib/discordrb/webhooks/modal.rb
Droid00000: Going to tentatively say this is resolved as well, see https://discord.com/channels/81384788765712384/381891448884428801/1441982914563670066

PR #298, opened 2025-03-10
GitHub Workflows & Rubocop

Droid00000:
# Summary

- Rubocop seems to have introduced a new plugins system.
- The current Github workflows are running on a [deprecated version](https://github.blog/changelog/2025-01-10-code-scanning-codeql-action-v2-is-now-deprecated/).

### Rubocop

Rubocop seems to have migrated over to a new [plugin system](https://docs.rubocop.org/rubocop/plugin_migration_guide.html).

- All this seems to require on our end is changed the `require:` key to `plugins:` in the yaml file.

`Lint/SuppressedExceptionInNumberConversion`
- I'm guessing this should mainly be a safe change to make. Setting `exception: false` seems to behave identically to suppressing the exception with a begin-rescue block.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -33,7 +33,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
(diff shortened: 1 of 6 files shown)


Droid00000:
Merged via #304

PR #302, opened 2025-03-24
style: satisfy all rubocop complaints

ibramsterdam:
# Summary
Goal of the pr is to satisfy all Rubocop rules so that for each future pr the ci is not giving false negatives

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -109,7 +109,7 @@ module Discordrb::Commands
         spaces_allowed: attributes[:spaces_allowed].nil? ? false : attributes[:spaces_allowed],
 
         # Webhooks allowed to trigger commands
-        webhook_commands: attributes[:webhook_commands].nil? ? true : attributes[:webhook_commands],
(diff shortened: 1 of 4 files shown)


lib/discordrb/commands/command_bot.rb
```diff
@@ -153,7 +153,9 @@ def initialize(**attributes)
             command = command.aliased_command
             command_name = command.name
           end
+          # rubocop:disable Lint/ReturnInVoidContext
           return "The command `#{command_name}` does not exist!" unless command
```
ibramsterdam, line 157: To scared to ask why the initializer is returning a string anyways ^-^

swarley: This is the registration for the default help command, so the return here in theory should be a response to the command

ibramsterdam: I took another look and I now see why we're doing it this way now @swarley . In the future, it might be a good idea to move this logic to a private method, like `register_help_command` or `setup_help_command`, and then just call that method from the initializer.

That being said, if you could review the rest of the PR when you have a chance, that would be awesome! 😊

ibramsterdam:
Closing PR since its probably outdated

PR #303, opened 2025-03-31
feat: missing role attributes

Droid00000:
## Summary

This PR adds support for unicode emojis as role icons and the `flags` field for server roles

## Added

`Role#flags`

`Role#unicode_emoji` and `Role#unicode_emoji=`

`Role#display_icon` A way to return the absolute icon a role has displayed in the Discord client

`Role#display_icon=` This would be a "smart" way of setting the icon a role has displayed. I feel a bit iffy about this, but I can also see how this might be helpful


lib/discordrb/data/role.rb
```diff
@@ -48,16 +54,21 @@ class Tags
...
       # @return [true, false] Whether this role is available for purchase
       attr_reader :available_for_purchase
+      alias_method :available_for_purchase?, :available_for_purchase
 
       # @return [true, false] Whether this role is a guild's linked role
       attr_reader :guild_connections
+      alias_method :guild_connections?, :guild_connections
+      alias_method :server_connections?, :guild_connections
 
+      # @!visibility private
```
Daniel-Worrall, line 71: Why?

lib/discordrb/data/role.rb
```diff
@@ -183,6 +209,28 @@ def icon_url(format = 'webp')
...
+      icon_url || unicode_emoji
+    end
+
+    # Set the icon this role is displaying.
+    # @param icon [File, String, nil] File like object that responds to #read, unicode emoji, or nil.
+    # @note Setting the icon to nil will remove the unicode emoji **and** the custom icon.
+    def display_icon=(icon)
+      # rubocop:disable Lint/ReturnInVoidContext
+      return update_role_data(unicode_emoji: nil, icon: nil) if icon.nil?
+      # rubocop:enable Lint/ReturnInVoidContext
```
Daniel-Worrall: I don't want to disable per line here, this could just as easily be reworked to a block that acts and then returns for more clarity that the return value isn't usable data

Also, if `icon` is nil, the below else block should run.
```
irb(main):001> nil.read
(irb):1:in `<main>': undefined method `read' for nil (NoMethodError)
```

Was this copied from somewhere else in the codebase?

Daniel-Worrall (changes requested):

lib/discordrb/data/role.rb
swarley: Does a plain `@private` not achieve what we want? I assume the intention is to hide the initializer from the docs

swarley: boomer ruby strikes again, this looks fine

swarley (approved):

PR #305, opened 2025-04-27
bugfixes: emoji parsing & removing timeouts

Droid00000:
## Summary
In the past couple of months, I've noticed a few small bugs related to parsing emojis, removing timeouts so I thought I'd take a crack at fixing them.

### Emoji Parsing
When parsing an emoji that isn't accessible via `Bot#emoji`, we previously attempted to do this:

```ruby
{ 'animated' => !animated.nil? }
```

Which won't work properly, because animated will always be some kind of string never nil, as pointed out by @Dakurei [here on Discord](https://discord.com/channels/81384788765712384/381891448884428801/1276916192849956927).

--- 
### Change Bot's Nickname
When trying to change the bot's own nickname we use the `guilds/#{server_id}/members/@me/nick` endpoint. We should probably swap to `guilds/#{server_id}/members/@me`, since the latter is what Discord seems to prefer, shown by the [deprecation warning](https://discord.com/developers/docs/resources/guild#modify-current-user-nick).

Additionally, the library previously converted `nil` to an empty string to signify the removal of a nickname. It seems like Discord has finally update the API calls to support sending `nil` since the `nick` parameter is marked as being nullable.

---
### Removing Timeouts
Previously when trying to remove a member's timeout by passing in nil, it would yield the following error:

```ruby
undefined method `iso8601' for nil
```

This is because `time` was always assumed to be a time object. We can solve this pretty easily by chucking in a  safe nav `&` before calling `#iso8601`.

---
### Wrong Return for `#search_members`
This last one was caused by me a few months ago 💀

```ruby
def search_members(name:, limit: nil)
  response = JSON.parse(API::Server.search_guild_members(@bot.token, @id, name, limit))
  return nil if response.empty?

  response.map { |mem| Member.new(mem, self, @bot) }
end
 ```
 
We probably shouldn't return `nil` here, since it's kinda inconsistent with how the lib handles other endpoints that return an array of data. I also didn't document `nil` as a valid return type for this method, so I made it simply return the empty array instead.

### DIFF
--- a/lib/discordrb/data/server.rb
+++ b/lib/discordrb/data/server.rb
@@ -588,12 +588,12 @@ module Discordrb
     # Searches a server for members that matches a username or a nickname.
     # @param name [String] The username or nickname to search for.
     # @param limit [Integer] The maximum number of members between 1-1000 to return. Returns 1 member by default.
-    # @return [Array<Member>, nil] An array of member objects that match the given parameters.
+    # @return [Array<Member>, nil] An array of member objects that match the given parameters, or nil for no members.
     def search_members(name:, limit: nil)
-      response = API::Server.search_guild_members(@bot.token, @id, name, limit)
+      response = JSON.parse(API::Server.search_guild_members(@bot.token, @id, name, limit))
       return nil if response.empty?
- 
-      JSON.parse(response).map { |member| Member.new(member, self, @bot) }
+
+      response.map { |mem| Member.new(mem, self, @bot) }
     end
 
     # Retrieve banned users from this server.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/discordrb/data/server.rb [hunk 1]
```diff
@@ -590,10 +590,8 @@ def max_emoji
     # @param limit [Integer] The maximum number of members between 1-1000 to return. Returns 1 member by default.
     # @return [Array<Member>] An array of member objects that match the given parameters.
     def search_members(name:, limit: nil)
-      response = JSON.parse(API::Server.search_guild_members(@bot.token, @id, name, limit))
-      return nil if response.empty?
```
Daniel-Worrall, line 594: We're changing the method signature here so I don't want this getting merged in until 4.0, can we split this out?

Daniel-Worrall (changes requested):
Approve except 1 undesired change at this time

Daniel-Worrall (approved):

PR #308, opened 2025-05-09
feat: resume gateway URL

Droid00000:
## Summary
This PR adds support for targeted switching, aka using the `resume_gateway_url` when resuming a gateway connection.

## Added
- `resume_gateway_url` argument to the initializer of `Session`.
-   Added the following line to `Session#invalidate`: `@resume_gateway_url = nil`

## Changed
- Make `Gateway#process_gateway` attempt to use the resume URL first, and then fall back to querying the Gateway URL via REST. E.g.

```ruby
 raw_url = @session&.resume_gateway_url || find_gateway
```

- I hope its alright that I added a bit of documentation for a few of the fields on the `Session` object

### DIFF
--- a/lib/discordrb/gateway.rb
+++ b/lib/discordrb/gateway.rb
@@ -93,7 +93,7 @@ module Discordrb
     # @return [String] Used to uniquely identify this session. Mostly used when resuming connections.
     attr_reader :session_id
 
-    # @return [Integer] Incrementing integer used to detemrine the most recent event reccived from Discord.
+    # @return [Integer] Incrementing integer used to determine the most recent event reccived from Discord.
     attr_accessor :sequence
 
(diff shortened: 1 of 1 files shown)


lib/discordrb/gateway.rb
```diff
@@ -656,7 +665,9 @@ def handle_error(e)
       LOGGER.log_exception(e)
     end
 
+    # rubocop:disable Lint/UselessConstantScoping
     ZLIB_SUFFIX = "\x00\x00\xFF\xFF".b.freeze
+    # rubocop:enable Lint/UselessConstantScoping
```
swarley, line 670: rubocop is maybe drunk

swarley (approved):

PR #310, opened 2025-05-26
fix: allow raw and application command event handlers to co-exist

Droid00000:
## Summary
This PR fixes an issue where registering application command event handlers and a raw event handler would cause errors like these:

```ruby
#<Thread:0x0000000145776960 /opt/homebrew/lib/ruby/gems/3.4.0/bundler/gems/discordrb-74d6415c6e15/lib/discordrb/bot.rb:1574 run> terminated with exception (report_on_exception is true):
/opt/homebrew/lib/ruby/gems/3.4.0/bundler/gems/discordrb-74d6415c6e15/lib/discordrb/bot.rb:1575:in 'block in Discordrb::Bot#handle_dispatch': undefined method 'interaction' for an instance of Discordrb::Events::RawEvent (NoMethodError)

            Thread.current[:discordrb_name] = "it-#{event.interaction.id}"
``` 

## Fixed

Now please hear me out, but the fix to this was actually changing the name of `event` to `cmd_event`.

Application commands execute asynchronously within a thread, which means Ruby doesn't need to wait for anything in that thread to finish running and it can continue running the main program, which in our case, would be checking for a `RawEvent`:

```ruby
if @event_handlers && @event_handlers[RawEvent]
  event = RawEvent.new(type, data, self) # This reassignment here is the issue
  raise_event(event)
end
```

But, threads may not always start running as soon as they are created, i.e. the VM lock might prevent the thread from immediately running (or it could just be that PCs are very fast). Since `event` gets re-assigned right away to `RawEvent` if a handler is registered for `RawEvent`, the `event` is now a `RawEvent`, and by the time the application command thread starts executing, `event` has been re-assigned.

The simplest solution is to use a separate variable name, in order to prevent the problem of `event (ApplicationCommandEvent)` getting re-assigned to `event (RawEvent)`

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -1571,13 +1571,13 @@ module Discordrb
         when Interaction::TYPES[:command]
           event = ApplicationCommandEvent.new(data, self)
 
-          Thread.new do
-            Thread.current[:discordrb_name] = "it-#{event.interaction.id}"
+          Thread.new(event) do |evt|
+            Thread.current[:discordrb_name] = "it-#{evt.interaction.id}"
 
             begin
-              debug("Executing application command #{event.command_name}:#{event.command_id}")
+              debug("Executing application command #{evt.command_name}:#{evt.command_id}")
 
-              @application_commands[event.command_name]&.call(event)
+              @application_commands[evt.command_name]&.call(evt)
             rescue StandardError => e
               log_exception(e)
             end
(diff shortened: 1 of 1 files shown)


Daniel-Worrall (changes requested):
I'd like to fix this properly. This is not the only place an assignment can occur that will cause problems for the MT code

Let's pass in the event object to the thread so the data is guaranteed at time of thread creation as a local variable
```
Thread.new(event) do |evt|
  ... evt
```
Let's do this both for `when Interaction::TYPES[:command]` and `call_event`

Droid00000:
> I'd like to fix this properly. This is not the only place an assignment can occur that will cause problems for the MT code
> 
> Let's pass in the event object to the thread so the data is guaranteed at time of thread creation as a local variable
> 
> ```
> Thread.new(event) do |evt|
>   ... evt
> ```
> 
> Let's do this both for `when Interaction::TYPES[:command]` and `call_event`

Alrighty, made the changes 👍

Daniel-Worrall (approved):

PR #313, opened 2025-06-14
feat: missing member & user attributes

Droid00000:
## Summary

This pull request adds in missing properties for server members and base users. 

### New Base Classes
`Discordrb::AvatarDecoration`
`Discordrb::Collectibles`
`Discordrb::Collectibles::Nameplate`

### New CDN Methods
`API#avatar_decoration_url`
`API#nameplate_url`

### New Member Properties
`MemberAttributes::MEMBER_FLAGS`
`MemberAttributes#flags`
`MemberAttributes#pending?`
`MemberAttributes#server_avatar_id`
`MemberAttributes#server_banner_id`
`MemberAttributes#server_avatar_decoration`

`Member#display_avatar`
`Member#display_banner`
`Member#display_avatar_decoration`
`Member#flags=`

`flags:` KWARG to `API::Server#update_member`

`API::Server#avatar_url` (from Swarley)
`API::Server#banner_url` (from Swarley)

### New User Properties
`UserAttributes#system_account?`
`UserAttributes#avatar_decoration`
`UserAttributes#collectibles`
`UserAttributes#banner_url`

`User#banner_id` Hear me out, but there's a reason this falls back to an API request. User banners are only returned when the user was directly fetched via an API call. Unlike when other properties on the base user change, `:GUILD_MEMBER_UPDATE` does **not** fire when changing your banner. So, this is the only way to reliably fetch the banner for a user.

`API::User#banner_url`

## Changed

When handling `:GUILD_MEMBER_UPDATE` update events, we attempt to unconditonally update most member attributes. This is fine for fields that are always present like `roles` but, some of these fields, e.g. `premium_since` and `timeout_until` are optional in the documentation.

The proper way to handle this would be to check for the existence of the keys themselves with `Hash#key?`. Funnily enough, `Member#update_data` was already correctly handling things. It's as if that method was meant to be used for `GUILD_MEMBER_UPDATE` events but someone forgot to plug the logic in.

I made the code delegate to `Member#update_data`, since it personally didn't make much sense to me to have basically the same code in two places.

Btw, I noticed that `:GUILD_MEMBER_UPDATE` is fired for changes to the underlying user, but we weren’t updating most user only fields. I added in methods to update a few commonly changing ones such as`update_avatar_decoration` and `update_collectibles`.


lib/discordrb/bot.rb
```diff
@@ -1102,11 +1102,11 @@ def update_guild_member(data)
...
       if (member = server.member(data['user']['id'].to_i))
-        member.update_roles(data['roles'])
-        member.update_nick(data['nick'])
+        member.update_data(data)
         member.update_global_name(data['user']['global_name']) if data['user']['global_name']
-        member.update_boosting_since(data['premium_since'])
-        member.update_communication_disabled_until(data['communication_disabled_until'])
+        member.avatar_id = data['user']['avatar'] if data['user'].key('avatar')
+        member.update_avatar_decoration(data['user']['avatar_decoration_data']) if data['user'].key?('avatar_decoration_data')
+        member.update_collectibles(data['user']['collectibles']) if data['user'].key?('collectibles')
```
swarley: I believe we talked about moving this all to the member's update_data function. I see we're adding some key presence guards to it already, should we just throw this logic into the method instead?

lib/discordrb/data/user.rb
```diff
@@ -167,6 +188,12 @@ def send_file(file, caption = nil, filename: nil, spoiler: nil)
       pm.send_file(file, caption: caption, filename: filename, spoiler: spoiler)
     end
 
+    # @return [String, nil] the ID of this user's current banner, can be used to generate a banner URL.
+    # @see #banner_url
+    def banner_id
+      @banner_id ||= JSON.parse(API::User.resolve(@bot.token, @id))['banner']
```
swarley, line 194: This feels a bit strange. We're in the context of a user so are we guarding against the case that we have an out of date cache? If these entities are updated via the gateway we should consider not doing a separate request here

lib/discordrb/data/user.rb
```diff
@@ -132,6 +149,10 @@ def initialize(data, bot)
 
       @status = :offline
       @client_status = process_client_status(data['client_status'])
+      @banner_id = data['banner']
+      @system_account = data.key?('system') ? data['system'] : false
```
swarley: This key based guard seems somewhat pointless during the saturation step. I would maybe just do `data['system'] || false` or `!!data['system']`

lib/discordrb/data/user.rb
swarley: Okay discord is just insane and I guess we just don't have the technology for this to be exposed on the gateway

swarley (approved):
Looks good, appreciate the updates

PR #320, opened 2025-06-26
fix: errors on :THREAD_LIST_SYNC

Droid00000:
# Summary
When handling the `:THREAD_LIST_SYNC` event, we incorrectly pass the guild_id directly into `ensure_channel`. `ensure_channel` expects a full server object. This is fine if the threads are already cached. However, the line [here](https://github.com/Droid00000/discordrb/blob/58afd8c360ec9fe5a97c45221a7510d68fd8df9f/lib/discordrb/data/channel.rb#L139) assumes that server is always a server object, which can cause issues for uncached threads

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -1652,7 +1652,8 @@ module Discordrb
 
         # raise ThreadDeleteEvent
       when :THREAD_LIST_SYNC
-        server = self.server(data['guild_id'].to_i)
+        server_id = data['guild_id'].to_i
+        server = @servers[server_id]
 
         # The `channel_ids` field has two meanings:
         #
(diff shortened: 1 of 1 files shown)


lib/discordrb/bot.rb
```diff
@@ -1652,8 +1652,10 @@ def handle_dispatch(type, data)
 
         # raise ThreadDeleteEvent
       when :THREAD_LIST_SYNC
+        server = self.server(data['guild_id'].to_i)
+
         data['members'].map { |member| ensure_thread_member(member) }
-        data['threads'].map { |channel| ensure_channel(channel, data['guild_id']) }
+        data['threads'].map { |channel| ensure_channel(channel, server) }
```
swarley: If we don't have a full server object we should just pass nil here, the fetching of the server will happen lazily in the channel

Droid00000: > If we don't have a full server object we should just pass nil here, the fetching of the server will happen lazily in the channel

If we're going to implement the other missing parts of `:THREAD_LIST_SYNC`, we're going to need to resolve the server anyways. Not too fussed about it, but would you still prefer the server gets lazily resolved?

swarley: I think that it's preferable to keep HTTP stuff out of the gateway processing loop if we can avoid it

PR #323, opened 2025-07-02
feat: server tags

Droid00000:
## Summary

This PR adds in server tags. Server tags can be a little wonky to work with because:

- The field itself can be `nil`
- The object can be present, but all the fields within it are `nil`/`false`
- All fields except `identity_guild_id` and `identity_enabled` can be `nil` (the server has changed their tag and the user needs to readopt it)

I think the easiest way to deal with this could be to not serialize the object at all if `identity_enabled` isn't set to `true`

## Added

`PrimaryServer` 
`User#primary_server` (aliased as `User#server_tag`)
`API#server_tag_badge_url`


lib/discordrb/data/primary_server.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module Discordrb
+  # A server tag that a user has chosen to display on their profile.
+  class PrimaryServer
```
swarley, line 5: every day we stray further from the `Guild` light

swarley (approved):

PR #326, opened 2025-07-12
feat: application command permission overwrites

Droid00000:
## Summary

This PR adds support for fetching application command permissions and listening for the `APPLICATION_COMMAND_PERMISSIONS_UPDATE` gateway event.

I wasn't really sure of the best way to handle this, since the object structure felt kinda weird to me, and it didn't seem like a good idea to nest the actual command overwrites. I'm also not a huge fan of using `Hash#except`, so suggestions are welcome lol

## Added

`ApplicationCommand::Permission`
`ApplicationCommand#permissions`
`ApplicationCommandPermissionsUpdateEvent`
`ApplicationCommandPermissionsUpdateEventHandler`
`Container#application_command_permissions_update`
`API::Application#get_application_command_permissions`


lib/discordrb/data/interaction.rb
```diff
@@ -398,6 +398,108 @@ def edit(name: nil, description: nil, default_permission: nil, &block)
...
+
+      # @!visibility private
+      def initialize(data, bot)
+        @bot = bot
+        @type = data['type']
+        @target_id = data['id'].to_i
+        @overwrite = data['permission']
+        @command_id = data['_command']['id'].to_i
+        @server_id = data['_command']['guild_id'].to_i
+        @application_id = data['_command']['application_id'].to_i
```
swarley: It looks like this is ephemeral (based on line 409), I would prefer to pass this as a second argument instead of embeding it into the data. `def initialize(data, command, bot)`

lib/discordrb/events/interactions.rb
```diff
@@ -533,4 +533,49 @@ def respond(choices:)
...
+    attr_reader :application_id
+
+    # @!visibility private
+    def initialize(data, bot)
+      @bot = bot
+      @server_id = data['guild_id'].to_i
+      @application_id = data['application_id'].to_i
+      @command_id = data['id'].to_i if data['id'].to_i != @application_id
+      @permissions = data['permissions'].map do |permission|
+        Discordrb::ApplicationCommand::Permission.new(permission.merge({ '_command' => data.except('permissions') }), bot)
```
swarley: same here as the previous comment

swarley (approved):

PR #340, opened 2025-08-22
feat: message forwards

Droid00000:
## Summary

Adds support for sending message forwards and receiving message snapshots

## Added
`Message#forward`
`Mesaage#snapshots`
`Discordrb::Snapshot`
`Message#to_reference`

## Fixed

Fixed a bug that caused allowed mentions to always be ignored when editing a message.

### Broken
```ruby
{ mentions: mentions, **rest }
```

### Fixed
```ruby
{ allowed_mentions: mentions, **rest }
```


lib/discordrb/data/attachment.rb
```diff
@@ -5,7 +5,7 @@ module Discordrb
   class Attachment
     include IDObject
 
-    # @return [Message] the message this attachment belongs to.
+    # @return [Message, Snapshot] the message or message snapshot this attachment belongs to.
```
swarley: This type change is a bit spooky and breaking.

swarley: Same for the embeds

Droid00000: I added the methods `#message` and `#snapshot` which returns the object if the type matches and removed the `attr_reader :message`, since putting the logic in the initializer didn't feel right for some reason

PR #343, opened 2025-08-30
Basic stickers support

### LINKED ISSUE #18: Sticker support
Message objects can have an array of [stickers](https://discord.com/developers/docs/resources/channel#message-object-message-sticker-structure) associated with them.

This will require a new class (`Sticker`) and an additional field (`#stickers`) on `Message`.

edshot99:
# Summary

Include sticker objects if stickers were sent in a message. Fixes #18

This could be expanded to let the bot access server sticker's, but I just wanted something basic at the moment.

Example
```
  event.message.stickers.each do |sticker|
    file_url = "https://cdn.discordapp.com/stickers/#{sticker.id}.#{sticker.extension}"
    populate_cdn_sticker("", file_url)
  end
```

---

## Added

- Sticker data type.

## Changed

- Message data type to include :sticker_items and a :stickers alias.

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -233,7 +233,7 @@ module Discordrb
       @pinned_at = data['pinned_at'] ? Time.parse(data['pinned_at']) : nil
 
       @sticker_items = []
-      @sticker_items = data['sticker_items'].map { |e| Sticker.new(e, self, @bot) } if data['sticker_items']
+      @sticker_items = data['sticker_items'].map { |e| Sticker::Item.new(e, self, @bot) } if data['sticker_items']
 
       @call = data['call'] ? Call.new(data['call'], @bot) : nil
 
--- a/lib/discordrb/data/sticker.rb
+++ b/lib/discordrb/data/sticker.rb
@@ -3,8 +3,6 @@
 module Discordrb
   # Server sticker
   class Sticker
-    include IDObject
-
     FORMATS = {
       png: 1,
       apng: 2,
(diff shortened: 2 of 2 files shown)


lib/discordrb/data/sticker.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module Discordrb
+  # Server sticker
+  class Sticker
```
Droid00000, line 5: How would you feel about doing something like the following for the time-being?

```ruby
class Sticker
  class Item
  ```
  
Since we probably wanna support full sticker objects as well, later down the road

File: lib/discordrb/data/sticker.rb [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+        @extension = 'png'
+      when 3
+        @extension = 'json'
+      when 4
+        @extension = 'gif'
+      end
+    end
+
+    # ID or name based comparison
+    def ==(other)
```
Droid00000: I think it might be okay to remove this method here entirely? Since all stickers will always have an ID, I think it should be alright to just use the `IDObject` [method](https://github.com/shardlab/discordrb/blob/cfce8ce5a5734fccaf40f0f4fa61d9b628cb01b8/lib/discordrb/id_object.rb#L11)

lib/discordrb/data/sticker.rb
```diff
@@ -0,0 +1,53 @@
...
+    def initialize(data, message, bot)
+      @bot = bot
+      @message = message
+
+      @name = data['name']
+      @description = data['description']
+      @id = data['id']&.to_i
+      @format_type = data['format_type']
+
+      case @format_type
```
Droid00000: Would you be open to making an enum for this, and adding meta-methods for this as well? E.g.

```ruby
FORMATS = {
  png: 1,
  apng: 2,
  # and so on...
}.freeze
```

```ruby
# @!method png?
#   @return [true, false]
# @!method ...
FORMATS.each do |name, value|
  define_method("#{name}?") do
    @format_type == value
  end
end

File: lib/discordrb/data/message.rb [hunk 4]
```diff
@@ -32,6 +32,10 @@ class Message
     # @return [Array<Embed>] the embed objects contained in this message.
     attr_reader :embeds
 
+    # @return [Array<Sticker>] the sticker objects contained in this message.
+    attr_reader :sticker_items
```
Droid00000, line 105: I think it's probably okay to just use `stickers` for this

Droid00000 (commented):
Thanks for the PR! Everything looks fine to me besides the comments

lib/discordrb/data/sticker.rb
edshot99: I'm not too familiar with the code base, but I thought a Sticker class is fine on its own.
If the Item class is needed then I think Sticker inheriting the Item class would be better.

lib/discordrb/data/sticker.rb
edshot99: Done.

lib/discordrb/data/sticker.rb
edshot99: Done.

lib/discordrb/data/message.rb
edshot99: In the original issue #18 [swarley](https://github.com/swarley) had said to use both.

Droid00000: Ah okay. If Swarley said that, it's cool with me.

lib/discordrb/data/sticker.rb
Droid00000: Please make it a seperate class. I.e. `Sticker::Item` for now. The reason for this is that full sticker objects have a lot more fields than a sticker Item, and thus we shouldn't represent them as the same thing

lib/discordrb/data/message.rb
```diff
@@ -228,6 +232,9 @@ def initialize(data, bot)
 
       @pinned_at = data['pinned_at'] ? Time.parse(data['pinned_at']) : nil
 
+      @sticker_items = []
```
Droid00000, line 235: It would be preferable if we used the safe navigation operator here and `OR` logic. It also isn't really necessary to provide the message to the sticker item. Nitpicky though

```ruby
(data['sticker_items'] || data['stickers'])&.map { |item| Sticker::Item.new(item, @bot) } || []
```

PR #344, opened 2025-09-01
add thread channel parmaeters to Channel#update

mrhappyma:
# Summary

---

## Added
parameters for thread channels to the channel update method, per API docs https://discord.com/developers/docs/resources/channel#modify-channel


mrhappyma:
can we move to non-position-dependant (idk the right word) arguments for non-required stuff? maybe i'm missing something but with the current pattern of just adding more and more positional stuff at the end, if i just want to unlock a thread i have to put in 12 `nil`s
```     Discordrb::API::Channel.update token, thread_id, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, false```
would be so much better if i could just do
```     Discordrb::API::Channel.update token, thread_id, locked: false```

would be one hell of a breaking change though

PR #345, opened 2025-09-01
docs: hide all initializers

Droid00000:
## Summary

This hides all the initializers from the docs. We already do this for most classes, but there were a few that were missing the `# @!visibility private` tag (mostly events), so I added them.


swarley:
Great change, hopefully it'll save some headache

swarley (approved):

PR #346, opened 2025-09-02
feat: base64 encoder

Droid00000:
## Summary

A few months ago in November, we [discussed](<https://discord.com/channels/81384788765712384/381891448884428801/1301199422444666922>) adding a Base64 encoding helper. This adds that, since base64 encoding is also required for a bunch of things that we don't support yet

## Added
`Discordrb.encode64`

### DIFF
--- a/lib/discordrb.rb
+++ b/lib/discordrb.rb
@@ -120,6 +120,19 @@ module Discordrb
       "<t:#{time.to_i}:#{TIMESTAMP_STYLES[style] || style}>"
     end
   end
+
+  # A utility method to base64 encode a file like object using its mime type.
+  # @param file [File, #read] A file like object that responds to #read.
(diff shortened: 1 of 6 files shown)


lib/discordrb.rb
```diff
@@ -120,6 +120,19 @@ def self.timestamp(time, style = nil)
       "<t:#{time.to_i}:#{TIMESTAMP_STYLES[style] || style}>"
     end
   end
+
+  # A utility method to base64 encode a file like object using its mime type.
+  # @param file [File, #read] A file like object that responds to #read.
+  # @return [String] The file object encoded as base64 image data.
+  def self.encode64(file)
```
swarley, line 127: I wonder if we should consider slightly changing this name to better indicate that it's for file data. Open to pushback

PR #350, opened 2025-09-11
feat: per-server avatars, banners, and bio for bots

Droid00000:
# Summary

Bots can now set per-server avatars, banners, and bio exclusively via the API.

I took this chance to add `Profile#banner=` and remove a few more user-account methods from `Profile`.

## Added
`Member#server_bio=`
`Member#server_avatar=`
`Member#server_banner=`
`API::User#update_current_user!`

## Fixed
`Profile#avatar_url` would sometimes be a base64 encoded string because the `avatar` would be set to the base64 encoded string on updates.


swarley:
We may want to expose a method on the member to get their avatar with logic like this `server_avatar || avatar || default_avatar`

PR #352, opened 2025-09-14
fix: lazily resolve resolved message channels

Droid00000:
## Summary

Currently, if you run a message context command where the bot is not authorized with the `bot` scope, an error occurs because the bot can't access the channel for the message. Luckily, Discord provides a decent amount of data for a partial channel which we can use instead

## Changed
`Message#server` is now lazily resolved
`Message#role_mentions` is now lazily resolved

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -147,18 +147,13 @@ module Discordrb
       @bot = bot
       @id = data['id'].to_i
       @content = data['content']
+      @channel = bot.channel(data['channel_id'].to_i)
       @pinned = data['pinned']
       @type = data['type']
       @tts = data['tts']
       @nonce = data['nonce']
       @mention_everyone = data['mention_everyone']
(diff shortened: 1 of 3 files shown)


lib/discordrb/data/message.rb
```diff
@@ -151,21 +145,24 @@ class Message
...
       @tts = data['tts']
       @nonce = data['nonce']
       @mention_everyone = data['mention_everyone']
+      @webhook_id = data['webhook_id']&.to_i
+
+      @channel = if data['_channel_data']
+                   @bot.ensure_channel(data['_channel_data'])
+                 else
+                   bot.channel(data['channel_id'].to_i)
+                 end
```
swarley: This pattern feels a bit off, if we have the channel data before we get here we should have already ensured it and can do `bot.channel(...)` to retrieve it. Probably better to do that and then do a memoized function like you've done for server

```ruby
def channel
  @channel ||= bot.channel(@channel_id)
end
```

PR #357, opened 2025-09-22
feat: handle USER_UPDATE events

Droid00000:
## Summary

Discord has a [USER_UPDATE](https://discord.com/developers/docs/events/gateway-events#user-update) event which is fired whenever info about the bot's current user changes. You'd think that we'd already be covered by the `:GUILD_MEMBER_UPDATE` event for this case, but I've noticed that `:GUILD_MEMBER_UPDATE` doesn't fire whenever properties about the current bot change without the `:server_members` intent.

## Added
```ruby
when :USER_UPDATE
  @profile = Profile.new(data, self)
```


swarley (approved):
Looks good, I wonder if we care to make an event out of this as well. Can do this separately

PR #359, opened 2025-09-24
fix: server boost caching

Droid00000:
## Summary

I think `GUILD_UPDATE` is fired whenever someone boosts a server, or it loses a boost. Unfortunately, I don't think we have anything that updates this in the cache, which is what's probably causing the cache to become stale.

- So while I was going through the initializer, I saw that we actually double cache channels on `:GUILD_CREATE`. This is because the initializer called `#process_channels` and then `#update_data` would also call that method again. I think it's safe to say that we probably only should call this once.

- Another thing is that the server's owner can change, and bots will get a `:GUILD_UPDATE` when that happens. Unfortunately, this means we can't memoize the owner, since there's a possibility that the memoized owner isn't the same as the `@owner_id` variable. I learned this the hard way when someone transferred ownership of a server my bot was in and permissions were getting incorrectly calculated, since [`PermissionCalculator#permission?`](https://github.com/shardlab/discordrb/blob/0752e573194cd038346fbf6ec237b316f7c9ac8d/lib/discordrb/permissions.rb#L165) checks for ownership.

### DIFF
--- a/lib/discordrb/data/server.rb
+++ b/lib/discordrb/data/server.rb
@@ -64,25 +64,20 @@ module Discordrb
     # @!visibility private
     def initialize(data, bot)
       @bot = bot
-      @owner_id = data['owner_id'].to_i
       @id = data['id'].to_i
       @members = {}
       @voice_states = {}
       @emoji = {}
 
-      process_channels(data['channels'])
       update_data(data)
 
       # Whether this server's members have been chunked (resolved using op 8 and GUILD_MEMBERS_CHUNK) yet
       @chunked = false
-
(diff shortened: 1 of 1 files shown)


lib/discordrb/data/server.rb
```diff
@@ -64,25 +64,20 @@ class Server
     # @!visibility private
     def initialize(data, bot)
       @bot = bot
-      @owner_id = data['owner_id'].to_i
       @id = data['id'].to_i
       @members = {}
       @voice_states = {}
       @emoji = {}
 
-      process_channels(data['channels'])
```
swarley, line 73: are we losing anything by removing this call?

PR #360, opened 2025-09-24
feat: allow creation of forum posts

Droid00000:
## Summary
This PR adds support for creating forum posts. This is pretty similar to creating a normal message, except you can specify tags on the forum post and pass an audit log reason as well.


Droid00000:
> I've tested this code locally and unfortunately, it does not work for me. I get invalid form body

Can you share the snippet you used to test this?

PR #362, opened 2025-09-28
fix: use response-data to update server members

Droid00000:
## Summary

This PR makes the cached data for member objects be immediately updated upon updating a member, via the updated member object returned by the REST-API.

## Fixed
I noticed that when using `#inspect`, all of the voice state attributes would always be `nil`. This is because `#inspect` was referencing the attributes as if they were instance variables, when they're actually methods.

### DIFF
--- a/lib/discordrb/data/member.rb
+++ b/lib/discordrb/data/member.rb
@@ -192,7 +192,7 @@ module Discordrb
     def communication_disabled_until=(timeout_until)
       raise ArgumentError, 'A time out cannot exceed 28 days' if timeout_until && timeout_until > (Time.now + 2_419_200)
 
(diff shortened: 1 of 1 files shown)


swarley:
This is a good change, however we are doing a lot of repeating for `update_data(JSON.parse(API::Server.update_member(...)))`, maybe we do a private update function to centralize this to a single call

Droid00000:
> This is a good change, however we are doing a lot of repeating for `update_data(JSON.parse(API::Server.update_member(...)))`, maybe we do a private update function to centralize this to a single call

Alright, I went and added `update_member_data` and `update_current_member_data` respectively, and updated everything

swarley:
Much cleaner, good change

swarley (approved):

Droid00000:
> Much cleaner, good change

In the spirit of the changes done here, I'm going to update https://github.com/shardlab/discordrb/pull/350 in a bit since this PR affects that one pretty heavily, lol

PR #363, opened 2025-10-01
feat: fix thread caching

Droid00000:
# Summary

This PR fixes a small issue where thread related fields never ever get updated on `THREAD_UPDATE`. This was because even though we call `Channel#update_from(other)`, there's nothing in the method body which actually updates the thread-related fields.

## Added
`Channel#last_pin_timestamp`
Missing predicate alias for `Channel#archived`
Missing predicate alias for `Channel#invitable`

## Fixed
Thread-specific attributes never getting updated

### DIFF
--- a/lib/discordrb/data/channel.rb
+++ b/lib/discordrb/data/channel.rb
@@ -998,7 +998,7 @@ module Discordrb
     # @note For internal use only
     # @!visibility private
     def process_last_pin_timestamp(time)
-      @last_pin_timestamp = time.is_a?(String) ? Time.parse(time) : time
+      @last_pin_timestamp = time ? Time.parse(time) : time
     end
 
     # Updates the cached data with new data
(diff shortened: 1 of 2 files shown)


lib/discordrb/data/channel.rb
```diff
@@ -979,6 +993,14 @@ def remove_recipient(recipient)
       @recipients.delete(recipient)
     end
 
+    # Set the last pin timestamp of a channel.
+    # @param time [String, nil] the time of the last pinned message in the channel
+    # @note For internal use only
+    # @!visibility private
+    def process_last_pin_timestamp(time)
+      @last_pin_timestamp = time.is_a?(String) ? Time.parse(time) : time
```
swarley: if we're going to do a ternary do we need to do a string check? `time ? Time.parse(time) : nil`. nitpicky though

lib/discordrb/data/channel.rb
```diff
@@ -994,6 +1016,7 @@ def update_data(new_data = nil)
       @parent_id = new_data[:parent_id] || new_data['parent_id'] || @parent_id
       process_permission_overwrites(new_data[:permission_overwrites] || new_data['permission_overwrites'])
       @rate_limit_per_user = new_data[:rate_limit_per_user] || new_data['rate_limit_per_user'] || @rate_limit_per_user
+      process_last_pin_timestamp(new_data['last_pin_timestamp']) if new_data['last_pin_timestamp']
```
swarley: I wonder what happens if there's only one pin and then we unpin, should we be checking for key presence rather than key value truthiness?

swarley (approved):

PR #367, opened 2025-10-12
feat: more missing channel fields

Droid00000:
## Summary

I was going through the docs for guilds and I saw that `Server#region` has been deprecated. Turns out the API literally returns a string that says `"deprecated"`, lol. The alternative is `Channel#voice_region`, so I went and implemented that, along with 2-3 other missing voice channel fields.

## Added
`Channel#flags`
`Channel#voice_region`
`Channel#last_message`
`Channel#last_message_id`
`Channel#video_quality_mode`
`Channel#total_message_count`

`Channel#process_last_message_id`
`Channel#increment_total_message_count`

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -1361,9 +1361,6 @@ module Discordrb
           raise_event(ChannelCreateEvent.new(message.channel, self))
         end
 
-        # Only thread channels have a count, so don't do anything for other types.
-        message.channel.increment_total_message_count if message.channel.thread?
-
         message.channel.process_last_message_id(message.id)
 
         event = MessageEvent.new(message, self)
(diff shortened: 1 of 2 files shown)


lib/discordrb/data/channel.rb
```diff
@@ -149,6 +166,7 @@ def initialize(data, bot, server = nil)
       @rate_limit_per_user = data['rate_limit_per_user'] || 0
       @message_count = data['message_count']
       @member_count = data['member_count']
+      @total_message_count = data['total_message_sent'] || 0
```
swarley: would prefer that this matched the naming from discord unless we've got a good reason

lib/discordrb/bot.rb
```diff
@@ -1358,6 +1361,11 @@ def handle_dispatch(type, data)
           raise_event(ChannelCreateEvent.new(message.channel, self))
         end
 
+        # Only thread channels have a count, so don't do anything for other types.
+        message.channel.increment_total_message_count if message.channel.thread?
```
swarley: do we not get a count on the thread object that's delivered with the message?

swarley (approved):

PR #369, opened 2025-10-16
feat: gateway timeout

UlyssesZh:
# Summary

While Discordrb has been working flawlessly on my machine, after I moved to another place, it starts seeing errors like tihs frequently:

```
[WARN : heartbeat @ 2025-10-15 20:45:43.756] Last heartbeat was not acked, so this is a zombie connection! Reconnecting
[ERROR : heartbeat @ 2025-10-15 20:45:43.756] The websocket connection has closed: (no information)
[ERROR : websocket @ 2025-10-15 21:00:53.264] An error occurred in the main websocket loop!
[ERROR : websocket @ 2025-10-15 21:00:53.264] Exception: #<Errno::ETIMEDOUT: Connection timed out>
[ERROR : websocket @ 2025-10-15 21:00:53.264] /usr/local/lib/ruby/3.4.0/openssl/buffering.rb:160:in 'OpenSSL::SSL::SSLSocket#sysread'
[ERROR : websocket @ 2025-10-15 21:00:53.264] /usr/local/lib/ruby/3.4.0/openssl/buffering.rb:160:in 'OpenSSL::Buffering#readpartial'
[ERROR : websocket @ 2025-10-15 21:00:53.264] /usr/local/bundle/gems/discordrb-3.5.0/lib/discordrb/gateway.rb:602:in 'Discordrb::Gateway#websocket_loop'
[ERROR : websocket @ 2025-10-15 21:00:53.264] /usr/local/bundle/gems/discordrb-3.5.0/lib/discordrb/gateway.rb:579:in 'Discordrb::Gateway#connect'
[ERROR : websocket @ 2025-10-15 21:00:53.264] /usr/local/bundle/gems/discordrb-3.5.0/lib/discordrb/gateway.rb:473:in 'block in Discordrb::Gateway#connect_loop'
[ERROR : websocket @ 2025-10-15 21:00:53.264] <internal:kernel>:168:in 'Kernel#loop'
[ERROR : websocket @ 2025-10-15 21:00:53.265] /usr/local/bundle/gems/discordrb-3.5.0/lib/discordrb/gateway.rb:472:in 'Discordrb::Gateway#connect_loop'
[ERROR : websocket @ 2025-10-15 21:00:53.265] /usr/local/bundle/gems/discordrb-3.5.0/lib/discordrb/gateway.rb:168:in 'block in Discordrb::Gateway#run_async'
[INFO : websocket @ 2025-10-15 21:00:53.265] Instant reconnection flag was set - reconnecting right away
[INFO : websocket @ 2025-10-15 21:00:54.008] Discord using gateway protocol version: 9, requested: 9
```

Notice that the time between the two errors is 15min 10s (and it always is whenever this happens), which is a pretty long time for a Discord bot to be down. This timeout threshold is probably set somewhere in `/proc/sys/net/ipv4/tcp_*`(and nowhere else sets a timeout, so it will otherwise just hang indefinitely), but I am not sure. Anyway, the workaround that I have come up with requires some modification to Discordrb which I have done in this PR. It should not affect any existing users, but will help in this particular case where the socket becomes zombie before the first hello message.

I am not sure whether the 2s initial timeout I set here is too short or not, but I think it will not normally be exceeded even in very poor network conditions. If it is too short, I can make it to be configurable by an environment variable so that people can set this according to the network condition.

## Added

Sets a timeout on the socket connected to Discord gateway.

### DIFF
--- a/lib/discordrb/gateway.rb
+++ b/lib/discordrb/gateway.rb
@@ -152,6 +152,11 @@ module Discordrb
     # - 4014: Use of disabled privileged intents.
     FATAL_CLOSE_CODES = [4003, 4004, 4011, 4014].freeze
 
+    # The timeout in seconds for responses from Discord when expecting one.
+    # This is used when expecting a hello message after the initial connection
+    # and when expecting a heartbeat ack after sending a heartbeat (if {#check_heartbeat_acks} is true).
+    RESPONSE_TIMEOUT = ENV['DISCORDRB_GATEWAY_TIMEOUT']&.to_f
+
     # Heartbeat ACKs are Discord's way of verifying on the client side whether the connection is still alive. If this is
     # set to true (default value) the gateway client will use that functionality to detect zombie connections and
     # reconnect in such a case; however it may lead to instability if there's some problem with the ACKs. If this occurs
@@ -279,6 +284,9 @@ module Discordrb
         end
 
         @last_heartbeat_acked = false
+
+        # A heartbeat ack should be received quickly after heartbeat
+        socket_timeout(0)
       end
 
       send_heartbeat(@session ? @session.sequence : 0)
@@ -451,9 +459,24 @@ module Discordrb
 
     private
 
+    def socket_timeout(duration)
+      return unless RESPONSE_TIMEOUT
+
+      duration += RESPONSE_TIMEOUT
+      if @socket.respond_to?(:timeout=)
+        @socket.timeout = duration
+        LOGGER.debug("Set socket timeout to #{duration}")
+      else # Ruby version < 3.4
+        LOGGER.debug("Socket timeout unsupported on Ruby < 3.4; skipping timeout configuration (current: #{RUBY_VERSION})")
+      end
+    end
+
     def setup_heartbeats(interval)
       # Make sure to reset ACK handling, so we don't keep reconnecting
       @last_heartbeat_acked = true
+
+      # Socket timeout needs to be larger than heartbeat interval
+      socket_timeout(interval + 1)
 
       # We don't want to have redundant heartbeat threads, so if one already exists, don't start a new one
       return if @heartbeat_thread
@@ -588,6 +611,9 @@ module Discordrb
       @socket = obtain_socket(gateway_uri)
       LOGGER.debug('Obtained socket')
 
+      # A hello message should be received quickly after it is connected
+      socket_timeout(0)
+
       # Initialise some properties
       @handshake = ::WebSocket::Handshake::Client.new(url: url) # Represents the handshake between us and the server
       @handshaked = false # Whether the handshake has finished yet
@@ -622,6 +648,10 @@ module Discordrb
           rescue EOFError
             @pipe_broken = true
             handle_internal_close('Socket EOF in websocket_loop')
+            next
+          rescue IO::TimeoutError
+            @pipe_broken = true
+            handle_internal_close('Socket did not receive a response in time')
             next
           end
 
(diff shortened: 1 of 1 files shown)


lib/discordrb/gateway.rb
```diff
@@ -803,7 +831,10 @@ def handle_hello(packet)
     # Op 11
     def handle_heartbeat_ack(packet)
       LOGGER.debug("Received heartbeat ack for packet: #{packet.inspect}")
-      @last_heartbeat_acked = true if @check_heartbeat_acks
+      return unless @check_heartbeat_acks
+
+      @last_heartbeat_acked = true
+      socket_timeout(@heartbeat_interval + 1)
```
Droid00000, line 840: Do we need to set the value again here?

lib/discordrb/gateway.rb
```diff
@@ -267,6 +272,9 @@ def inject_error(e)
     # @see #send_heartbeat
     def heartbeat
       if check_heartbeat_acks
+        # This check is still needed even though we have socket_timeout
```
Droid00000: I don't think this comment is really needed. We're always going to support the `check_heartbeat_acks` property

Droid00000 (commented):
Personally, I am a fan of this, but we should make this something that's opt-in. Like the `check_heartbeat_acks=` setter

lib/discordrb/gateway.rb
UlyssesZh: Yes, because it was overridden at `socket_timeout(RESPONSE_TIMEOUT)` in `heartbeat`.

UlyssesZh:
> Personally, I am a fan of this, but we should make this something that's opt-in. Like the `check_heartbeat_acks=` setter

Basically I rewrote the behavior of `check_heartbeat_acks`. The user opts out by setting it to false.

lib/discordrb/gateway.rb
```diff
@@ -453,10 +453,22 @@ def send_raw(data, type = :text)
 
     private
 
+    def set_socket_timeout(duration)
+      if @socket.respond_to?(:timeout=)
+        @socket.timeout = duration
+        LOGGER.debug("Set socket timeout to #{duration}")
+      else # Ruby version < 3.4
+        LOGGER.debug('Cannot set socket timeout because Ruby is too old')
```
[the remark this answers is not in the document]
Daniel-Worrall: valid qualm for the output

UlyssesZh:
Sorry for bothering, but do you think this is ready for merge?

Droid00000:
> Sorry for bothering, but do you think this is ready for merge?

There was a bit of discussion about this a few months ago: https://discord.com/channels/81384788765712384/381891448884428801/1438303845913591881

UlyssesZh:
I read the discussions, but I do not understand the confusion. If the only concern is that 2 s is too short, we can easily make it configurable. Actually, 2 s is much more than enough because latency is usually less than 200 ms even in poor conditions.

Droid00000:
From what I understand @swarley said the following is non-negotiable: 

1. The behavior should be opt-in be default. It shouldn't be something that you have to opt-out of. That means the Gateway should operate exactly as it did prior to this PR unless you explicitly opt-in to having a socket timeout.

2. ~~The duration of the timeout should be equivalent to 2 heartbeat intervals, and it should be non-configurable.~~

UlyssesZh:
What is the reason behind 2 heartbeat intervals as the timeout? In 99% occasions it responds in 200 ms, and among the rest 1% occasions, in 99% occasions it actually is a zombie that needs a reconnection.

If the timeout is two heartbeat intervals, because my apartment cuts all connections every 51 min, my bot will have 7% downtime (on average it takes 2.5 heartbeat intervals to discover a zombie, and then it takes 2 heartbeat intervals to discover a zombie waiting for the first hello message (I don't know why in my network the first `#connect` after a reset always gives a zombie, but it does happen every time), so 4.5 heartbeats in total). If the timeout is very short, the downtime is reduced to 1.5% (on average it takes 0.5 heartbeat intervals to discover a zombie, and then it immediately reconnects after negligible timeouts). If we have no timeout at all (i.e., without this PR), then my bot has 32% downtime (OS-level timeout for sockets, which is a bit more than 15 min).

UlyssesZh:
Opt in by setting env var `DISCORDRB_GATEWAY_TIMEOUT`.

PR #377, opened 2025-10-27
feat: role move interface

Droid00000:
# Summary

This PR adds a `Role#move` method. Currently we already have a `Role#sort_above` method, and this works fine, but it's a little limited at the moment, and it isn't the most intuitive thing to use, imo.

## Added
`Role#move`

## Fixed
The `#mentionable` field not getting updated on `:GUILD_ROLE_UPDATE`

# EDITS

I made a small change to .rubocop.yml, because of the following warning it outputted:

```ruby
.rubocop.yml: Style:/EndOfLine has the wrong namespace - replace it with Layout/EndOfLine
```

I can make this a separate PR if desired.

### DIFF
--- a/lib/discordrb/data/role.rb
+++ b/lib/discordrb/data/role.rb
@@ -309,17 +309,9 @@ module Discordrb
     # @param other [Role, String, Integer, nil] The role, or its ID, above which this role should be moved. If it is `nil`,
     #   the role will be moved above the @everyone role.
     # @return [Integer] the new position of this role
+    # @deprecated Please migrate to using {#move} with the `above` or `below` KWARGS.
     def sort_above(other = nil)
-      other = @server.role(other.resolve_id) if other
-      roles = @server.roles.sort_by(&:position)
-      roles.delete_at(@position)
-
-      index = other ? roles.index { |role| role.id == other.id } + 1 : 1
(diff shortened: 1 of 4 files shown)


lib/discordrb/data/role.rb
```diff
@@ -327,13 +331,72 @@ def delete(reason = nil)
...
+    # @param bottom [true, false, nil] Whether to move the roles to the bottom of the role list.
+    # @param top [true, false, nil] Whether to move this role to the highest possible position.
+    # @param above [Integer, String, Role, nil] The role that this role should be moved above.
+    # @param below [Integer, String, Role, nil] The role that this role should be moved below.
+    # @param offset [Integer, nil] The number of roles to offset the new position by. A positive number will
+    #   move the role above, and a negative number will move the role below. This parameter is relative and
+    #   calculated after the `bottom`, `top`, `above`, and `below` parameters.
+    # @param reason [String, nil] The audit log reason to show for moving the role.
+    # @return [Integer] the new position of the role.
+    def move(bottom: nil, top: nil, above: nil, below: nil, offset: 0, reason: nil)
```
swarley: top/bottom might be potentially problematic given that a bot cannot move roles above the highest order role that it has (even with admin)

PR #379, opened 2025-10-28
fix: update Member#add_role and Member#remove_role

### LINKED ISSUE #378: `add_role` doesn't work after upgrading to 3.6.1
# Summary

---

## Environment

**Ruby version:**

ruby 3.4.5 (2025-07-16 revision 20cda200d3) +PRISM [arm64-darwin25]

**Discordrb version:**

3.6.1

`Exception: #<NoMethodError: undefined method 'one?' for an instance of Integer>`

Seems to have been introduced in https://github.com/shardlab/discordrb/pull/312. This line specifically:
https://github.com/shardlab/discordrb/blob/1279a28aa84544b3ca5faaa1738d3f9e8498553c/lib/discordrb/data/member.rb#L232

I don't think it would ever work because `integer.one?` isn't valid.

swarley:
Closes #378

### DIFF
--- a/lib/discordrb/data/member.rb
+++ b/lib/discordrb/data/member.rb
@@ -229,7 +229,7 @@ module Discordrb
     def add_role(role, reason = nil)
       role_ids = role_id_array(role)
 
-      if role_ids.count.one?
+      if role_ids.one?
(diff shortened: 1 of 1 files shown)


Droid00000 (approved):
1

PR #382, opened 2025-11-02
style: message initializer

Droid00000:
## Summary

Currently, the message initializer always creates a default array for each field with an array type, and then attempts to create another array for each field when calling `#map`. Some of these fields aren't nullable and already contain an array. So I think it might be faster to perform a short circuit operation via `||`

To prove this, I did a small experiment using this [Ruby Benchmark](https://github.com/ruby/benchmark) library on a sample of around 1,000 messages. The time it took for the original message initializer to cook was `0.00042731986143187056` seconds on average, while the time it took for the new version was `0.0003482551963048497` seconds.

## Changed
Make the `#emojis` array lazily resolved.
Use the safe navigation operator when initializing messages.

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -82,9 +82,6 @@ module Discordrb
     # @return [Channel] the channel in which this message was sent.
     attr_reader :channel
 
-    # @return [Time] the timestamp at which this message was sent.
-    attr_reader :timestamp
-
     # @return [Time] the timestamp at which this message was edited. `nil` if the message was never edited.
     attr_reader :edited_timestamp
     alias_method :edit_timestamp, :edited_timestamp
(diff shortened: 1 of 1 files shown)


swarley (approved):

lib/discordrb/data/message.rb
```diff
@@ -177,56 +179,31 @@ def initialize(data, bot)
...
       @snapshots = data['message_snapshots']&.map { |snapshot| Snapshot.new(snapshot['message'], @bot) } || []
-
       @role_subscription = RoleSubscriptionData.new(data['role_subscription_data'], self, @bot) if data['role_subscription_data']
-
-      @position = data['position'] || 0
     end
 
+    # @deprecated Please migrate to using {#creation_time} instead.
+    alias_method :timestamp, :creation_time
+
```
swarley, line 206: actually, what's the rationale for this one?

swarley (approved):

PR #385, opened 2025-11-11
feat: add ability to send webhook messages to threads

DorianGrayCodes:
# Summary

## Changed

Only modified the webhook client, gave extra thread_id param that you can pass in that adds it as a query param to the request

### DIFF
--- a/lib/discordrb/webhooks/client.rb
+++ b/lib/discordrb/webhooks/client.rb
@@ -96,7 +96,7 @@ module Discordrb::Webhooks
 
       yield builder if block_given?
 
-      query = URI.encode_www_form({thread_id: }.compact)
+      query = URI.encode_www_form({ thread_id: }.compact)
       data = builder.to_json_hash.merge({ content: content, embeds: embeds, allowed_mentions: allowed_mentions }.compact)
       RestClient.patch(
         "#{@url}/messages/#{message_id}#{(query.empty? ? '' : "?#{query}")}",


lib/discordrb/webhooks/client.rb
```diff
@@ -38,7 +38,7 @@ def initialize(url: nil, id: nil, token: nil)
     #     end
     #   end
     # @return [RestClient::Response] the response returned by Discord.
-    def execute(builder = nil, wait = false, components = nil)
+    def execute(builder = nil, wait = false, components = nil, thread_id = nil)
```
Droid00000: I think we should make this a KWARG here, and stick with positional arguments for the internal `post_*` methods

lib/discordrb/webhooks/client.rb
```diff
@@ -117,14 +117,20 @@ def avatarise(avatar)
       end
     end
 
-    def post_json(builder, components, wait)
+    def post_json(builder, components, wait, thread_id: nil)
       data = builder.to_json_hash.merge({ components: components.to_a })
-      RestClient.post(@url + (wait ? '?wait=true' : ''), data.to_json, content_type: :json)
+      RestClient.post(
```
Droid00000: I think we should use the `URI.encode_www_form` utility here instead of manually concatenating the strings. E.g.

```ruby
def post_json(builder, components, wait, thread_id)
  query = URI.encode_www_form({ wait:, thread_id: }.compact)
  data = builder.to_json_hash.merge({ components: components.to_a })
  RestClient.post(@url + (query.empty? ? '' : "?#{query}"), data.to_json, content_type: :json)
end
```

lib/discordrb/webhooks/client.rb
```diff
@@ -117,14 +117,20 @@ def avatarise(avatar)
...
+        @url + (wait ? '?wait=true' : '') + (thread_id ? "?thread_id=#{thread_id}" : ''),
+        data.to_json, content_type: :json
+      )
     end
 
-    def post_multipart(builder, components, wait)
+    def post_multipart(builder, components, wait, thread_id: nil)
```
Droid00000: Same as line 122

Droid00000 (approved):
Thanks. I've tested this out, and everything works.

PR #386, opened 2025-11-12
feat: static nameplate URLs

Droid00000:
# Summary

Nameplates have another path, where they render a single static frame, instead of an animated `webm` video. `/asset.webm` vs. `/static.png`


swarley:
Looks good. might also be nice to have a `static_url` on the Collectible that does the same thing without an argument requirement

swarley (approved):

PR #400, opened 2025-11-29
feat: channel updates

### LINKED ISSUE #195: Feat: Forums
At first glance, seems simple enough to implement - https://discord.com/developers/docs/topics/threads#forums

Should just be the case of adding a new channel type and testing.

Droid00000:
## Summary

This PR is a pretty big change. It changes how channel data is stored by channels. All of these changes are kinda needed to support modifying thread attributes, which isn't implemented yet. Since there's a bunch of things happening, I'll try and list them in short bullet points.

## Added
- Setters for all of the attributes that didn't have setters were added.

- Introduction of `Channel#update!`. The current `Channel#update` method is essentially unusable IMHO, especially for implementing boolean fields for threads. I personally believe writing a bunch of boilerplate like the example below is not really a good idea.

 ```ruby
 data.key?(:locked) ? data[:locked] : :undef
 ```
 
 `Channel#default_auto_archive_duration`
 `Channel#default_sort_order`
 `Channel#default_forum_layout`
 `Channel#default_thread_rate_limit_per_user`

## Changed
- `Channel#update_channel_data` was refactored to use the new API method.

- `Bot#update_channel` was refactored to point to `Channel#update_data`.

## Removed
 - Removal of `Channel#update_from`. Currently, this method was only used by the gateway to update data. The only other place where we have the `#update_from` pattern is roles. In it's place, I've brought `Channel#update_data` up-to-date with the API, and re-factored it to account for all of the missing fields. My primary motivation for this is that I don't think it's very maintainable to have to update three different method (`#initialize`, `#from_other`, `#update_data`) every time a new field gets added.

- All of the code related to group DM gateway events was removed. This is because a bot will never receive gateway events for group DMs even though they can interact with group DMs to a limited capacity.

## Fixed
- `Channel#nsfw?` now correctly calculates the `nsfw` property based off of the `nsfw` property of its parent channel.

- `Channel#join_timestamp` was always previously `nil` because the IVAR was incorrectly named `@member_join` instead of `@join_timestamp`.

- Resolve #195

### DIFF
--- a/lib/discordrb/data/channel_tag.rb
+++ b/lib/discordrb/data/channel_tag.rb
@@ -45,14 +45,14 @@ module Discordrb
         **(Emoji.build_emoji_hash(emoji) if emoji != :undef)
       }.reject { |_, value| value == :undef }
 
-      @channel.__send__(:update_tags, to_h.merge(data), reason)
+      @channel.update_tags(to_h.merge(data), reason)
     end
 
     # Permenantly delete the channel tag.
     # @param reason [String, nil] The reason to show in the audit log for deleting the tag.
     # @return [nil]
     def delete(reason: nil)
-      @channel.__send__(:update_tags, { id: @id, d: true }, reason)
+      @channel.update_tags({ id: @id, d: true }, reason)
     end
 
     # @!visibility private
--- a/lib/discordrb/data/channel.rb
+++ b/lib/discordrb/data/channel.rb
@@ -1227,6 +1227,19 @@ module Discordrb
       @last_message_id = id
     end
 
+    # Set the available tags of a channel.
+    # @param tag [Hash] the data for the tag to create
+    # @param reason [String, nil] the reason to show in the audit log
+    # @note For internal use only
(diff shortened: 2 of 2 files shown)


lib/discordrb/data/channel_tag.rb
```diff
@@ -0,0 +1,69 @@
...
+    # @param reason [String, nil] The reason to show in the audit log for modifying the tag.
+    # @return [nil]
+    def modify(name: :undef, emoji: :undef, moderated: :undef, reason: nil)
+      data = {
+        name: name,
+        moderated: moderated,
+        **(Emoji.build_emoji_hash(emoji) if emoji != :undef)
+      }.reject { |_, value| value == :undef }
+
+      @channel.__send__(:update_tags, to_h.merge(data), reason)
```
swarley: might want to rethink this a bit or just make it yard protected visibility rather than an actual private method, send is usually a smell

swarley (changes requested):

PR #401, opened 2025-12-05
fix: fetching thread members

Droid00000:
## Summary

Currently, fetching thread members is 100% broken no matter what. This is because the library accidentally calls `#collect` on a hash with a block arity of one. This returns an array of two-element arrays with each key/value pair in the hash (see below). Additionally, fetching thread members requires making REST calls which is currently not handed at all, as the code incorrectly assumes that all thread members for a thread will be cached (which is never the case because the gateway never provides this data up-front except for the current thread member).

```ruby
irb(main):001> hash = { ruby: "stone", crystal: "rock" };

irb(main):002> hash.collect { |name| name }
=> [[:ruby, "stone"], [:crystal, "rock"]]
```

`API::Channel#list_thread_members` incorrectly accepts a `before` parameter when only `after` is actually supported by the API. I've made an effort to avoid making a breaking change, but due to the lack of documentation about how the method works, I've had to make a few assumptions. 

This is technically a breaking change, but leaving the `before` parameter is more dangerous than removing it. The API doesn't raise any error when `before` is passed, and thus when I was testing the method, it caused an infinite loop with the `Paginator`. I strongly believe that this sort of issue is better just cut-off from the root, rather than someone finding out from their bot hanging forever, but it's ultimately your call. 

## Added
`with_member:` argument to `API::Channel#list_thread_members`

## Fixed
`Channel#members` now fetches all of the members who have joined a thread.


lib/discordrb/api/channel.rb
```diff
@@ -574,8 +574,8 @@ def remove_thread_member(token, channel_id, user_id)
 
   # Get the members of a thread.
   # https://discord.com/developers/docs/resources/channel#list-thread-members
-  def list_thread_members(token, channel_id, before, limit)
```
swarley, line 577: This breaks semver

swarley (changes requested):

lib/discordrb/api/channel.rb
Droid00000: okay, fixed. I hope just making a bang method is fine, unless you'd prefer keeping the current one

lib/discordrb/data/channel.rb
```diff
@@ -1093,22 +1095,46 @@ def join_thread
...
     def members
-      @bot.thread_members[@id].collect { |id| @server_id ? @bot.member(@server_id, id) : @bot.user(id) }
+      return [] unless thread?
+
+      return (@bot.thread_members[@id]&.keys&.filter_map { |id| @bot.member(@server_id, id) } || []) if @thread_members_chunked
+
+      get_members = proc do |before = nil|
+        JSON.parse(API::Channel.list_thread_members!(@bot.token, @id, before, 100, true)).map do |data|
+          @bot.ensure_thread_member(data)
+          Member.new(data['member'], server, @bot).tap { |member| server&.cache_member(member) }
```
swarley, line 1114: maybe we should have a better helper to ensure members like this. non blocking for this pr

swarley (approved):

PR #412, opened 2026-01-02
feat: more voice state filters

Droid00000:
## Summary

This PR adds three new voice state filters `newly_joined`, `disconnected`, and `moved`. I'm not really sure about these names, so any better ideas would be much appreciated

## Added
`moved:` attribute filter for voice state events
`newly_joined:` attribute filter for voice state events
`disconnected:` attribute filter for voice state events


lib/discordrb/events/voice_state_update.rb
```diff
@@ -99,6 +99,30 @@ def matches?(event)
...
+        end,
+        matches_all(@attributes[:disconnected], event.channel) do |a, e|
+          case a
+          when TrueClass
+            e.nil?
+          when FalseClass
+            !e.nil?
+          end
+        end,
+        matches_all(@attributes[:moved], event) do |a, e|
```
gitnickolson, line 119:
```suggestion
        matches_all(@attributes[:channel_changed], event) do |a, e|
```

Since it's not always a "move" as in moving a member

lib/discordrb/events/voice_state_update.rb
```diff
@@ -99,6 +99,30 @@ def matches?(event)
                else
                  e
                end
+        end,
+        matches_all(@attributes[:newly_joined], event.old_channel) do |a, e|
```
gitnickolson:
```suggestion
        matches_all(@attributes[:joined], event.old_channel) do |a, e|
```

I would call it just "joined" to keep it aligned with "diconeected" (since you wouldn't say "newly disconnected" as well)

lib/discordrb/container.rb
```diff
@@ -297,6 +297,9 @@ def channel_recipient_remove(attributes = {}, &block)
     # @option attributes [true, false] :deaf Matches whether or not the user is deafened server-wide.
     # @option attributes [true, false] :self_mute Matches whether or not the user is muted by the bot.
     # @option attributes [true, false] :self_deaf Matches whether or not the user is deafened by the bot.
+    # @option attributes [true, false] :newly_joined Matches whether or not the user is newly connecting to voice.
```
Droid00000:
```suggestion
    # @option attributes [true, false] :connected Matches whether or not the user is newly connecting to voice.

File: lib/discordrb/events/voice_state_update.rb (hunk 1 above)
Comment (Daniel-Worrall) [reply lib/discordrb/events/voice_state_update.rb:119]: moved is prefered, and this is what it's called in permission descriptions - it is also only in the context of a voice state

File: lib/discordrb/events/voice_state_update.rb (hunk 2 above)
Comment (Daniel-Worrall) [reply lib/discordrb/events/voice_state_update.rb]: Should be `:connected`

Comment (Droid00000) [inline lib/discordrb/events/voice_state_update.rb]:
```suggestion
        matches_all(@attributes[:connected], event.old_channel) do |a, e|

File: lib/discordrb/events/voice_state_update.rb [hunk 4]
```diff
@@ -99,6 +99,30 @@ def matches?(event)
                else
                  e
                end
+        end,
+        matches_all(@attributes[:joined], event.old_channel) do |a, e|
```
Droid00000:
```suggestion
        matches_all(@attributes[:connected], event.old_channel) do |a, e|
```

PR #417, opened 2026-01-08
feat: optimize Bot#emoji

Droid00000:
## Summary

Usually, I don't like to make these kinds of changes but, for people running large bots in servers with large amount of emojis, this should help them out. Here's a [gist](https://gist.github.com/Droid00000/7066128603e4d3db99c8c9c8767df857) showing the differences so if anyone has any doubt, they can just see for themselves.

## Changed
Make `Bot#emoji` run faster when a bot is in lots of guilds with lots of emojis.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -199,12 +199,18 @@ module Discordrb
     #   The list of emoji the bot can use.
     #   @return [Array<Emoji>] the emoji available.
     def emoji(id = nil)
-      emoji_hash = servers.values.map(&:emoji).reduce(&:merge)
-      if id
-        id = id.resolve_id
-        emoji_hash[id]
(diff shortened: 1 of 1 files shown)


lib/discordrb/bot.rb
```diff
@@ -199,12 +199,18 @@ def thread_members
...
     #   @return [Array<Emoji>] the emoji available.
     def emoji(id = nil)
-      emoji_hash = servers.values.map(&:emoji).reduce(&:merge)
-      if id
-        id = id.resolve_id
-        emoji_hash[id]
+      if (id = id&.resolve_id)
+        @servers.each_value do |server|
+          emoji = server.emojis[id]
+          return emoji if emoji
```
swarley, line 205:
```suggestion
          server.emojis[id]
```

PR #421, opened 2026-01-18
feat: community invites

Droid00000:
## Summary

This API uses CSV files, but despite this, the actual formatting of the CSV is not validated or sanitized at all, and when fetching the `target_users`, you are returned the input file exactly as it was uploaded. Of course always requiring a file is not a good idea, which is why we can use `StringIO` to accept an array and then create a fake "File" to fool `RestClient`.

## Added
`Invite#flags`
`Invite#roles`
`Invite#target_users`
`Invite#target_users=`
`Channel#create_invite`
`Invite#target_users_worker`

`Invite#expires_at`
`Invite#stream_user`
`Invite#embedded_application`

`API::Channel#create_invite!`
`API::Invite#get_target_users`
`API::Invite#update_target_users`
`API::Invite#get_target_users_job_status`

## Deprecated
`Channel#make_invite`

## Removed
`Bot#accept_invite`
`API::Invite#accept`


Droid00000:
After a lot of consideration, this needs to be heavily re-worked and for some reason setting `target_users` fails whenever I try and use it in my test server (lol), meaning there's probably something wrong with my implementation. Since the API for this has had some changes since I initially wrote the code I'll re-factor this over the coming weeks and fix it

PR #422, opened 2026-01-19
feat: comparison operators for roles

Droid00000:
## Summary

This PR adds the `>`, `<`, `<=`,`>=`, and `<=>` operators for comparing roles via their position in the hierarchy. I understand this may be an unwanted change, but it makes comparing roles by their position a lot easier. Other libraries implement [similar interfaces](https://github.com/Rapptz/discord.py/blob/master/discord/role.py#L241) too. Additionally, by implementing the Darth Vader ship operator `<=>`, calling `#sort` on an `Array<Role>` now behaves correctly.

## Added
`Role#>`
`Role#<`
`Role#<=`
`Role#>=`
`Role#between?`

### DIFF
--- a/lib/discordrb/data/member.rb
+++ b/lib/discordrb/data/member.rb
@@ -257,42 +257,39 @@ module Discordrb
       end
     end
 
-    # @return [Role] the highest role this member has.
+    # Get the highest role this member has.
+    # @return [Role] The highest role this member has.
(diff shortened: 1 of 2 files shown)


swarley:
This one is a little fancy, let me stew on it a bit longer

swarley (approved):

PR #432, opened 2026-02-24
feat: Member#modify

### LINKED ISSUE #419: RFC - Support editing multiple attributes
Currently, we use a `setter=` to pattern for modifying attributes on classes. I would like to see support for a single monolithic method to modify multiple attributes in one go, e.g. `#modify`, `#edit`, etc. The only way to currently do this is by dropping down the `API::` methods, but that isn't great because you lose things like `Atom -> Int` resolution for enums, and base64 encoding. Additionally, `setter=` methods mean we can't support passing an audit log reason in most cases even though the API supports it.

Droid00000:
## Summary
Adds a `#modify` method to the `Member` object. One thing to discuss might be whether `avatar:`, `banner:`, and `bio:` should be split into their own separate method, e.g. `#modify_profile` or `#modify_me`. Closes #419 

## Added
`Member#modify`
`API::Server#update_current_member!`


lib/discordrb/api/server.rb
```diff
@@ -193,14 +148,20 @@ def update_member(token, server_id, user_id, nick: :undef, roles: :undef, mute:
   # Update the current member's properties.
   # https://discord.com/developers/docs/resources/guild#modify-current-member
   def update_current_member(token, server_id, nick = :undef, reason = nil, bio = :undef, banner = :undef, avatar = :undef)
```
swarley, line 195: should deprecate this one too

Droid00000: > should deprecate this one too

fixed

PR #439, opened 2026-03-15
fix: voice UDP causes main gateway to stop acking heartbeats

Droid00000:
## Summary

A while ago, I got an issue where after connecting to the VWS failed, the the VWS would not properly close and still attempted to send heartbeats. Additionally, because `@socket.recv(74)` reads from a socket and blocks, and we have other threads, this would cause the main Gateway thread to somehow stop receiving data too, because after the failed voice connection attempt, my bot wouldn't respond to any slash commands, and eventually I'd see a debug message stating that the last heartbeat was not acked, and then the code would just indefinitely hang. After much fucking around, adding a timeout to the `#receive_discovery_reply` seems to fix this, and none of the problems above seem to be an issue.

### DIFF
--- a/lib/discordrb/voice/network.rb
+++ b/lib/discordrb/voice/network.rb
@@ -70,10 +70,14 @@ module Discordrb::Voice
     # @return [Array(String, Integer)] the IP and port received from the discovery reply.
     def receive_discovery_reply
       # Wait for a UDP message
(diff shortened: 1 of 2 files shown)


Droid00000:
This is gonna be fixed by the DAVE PR anyways, so there's no point in duplicating this

PR #446, opened 2026-04-05
feat: missing voice-state attributes

Droid00000:
## Summary

This PR adds a lot of missing attributes to the `VoiceState` object. I also added some aliases for some existing predicate methods that return booleans, since they didn't have any before

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -1051,9 +1051,19 @@ module Discordrb
       server = server(server_id)
       return unless server
 
+      if (member_data = data['member'])
+        member = server.member(member_data['user']['id'].to_i, false)
+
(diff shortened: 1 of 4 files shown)


lib/discordrb/bot.rb
```diff
@@ -1051,9 +1051,19 @@ def update_voice_state(data)
       server = server(server_id)
       return unless server
 
+      if (member_data = data['member'])
+        member = server.member(member_data['user']['id'].to_i, false)
```
swarley: nitpick, this shouldn't need `#to_i` since it should work as a string. Doesn't matter either way really

Droid00000: fixed

PR #456, opened 2026-05-16
feat: Server#query_members

Droid00000:
## Summary

This PR adds a`#query_members` method to allow you to get multiple members by user ID in one go. It uses the `REQUEST_GUILD_MEMBERS` opcode with the `user_ids` parameter to do this. Example usage:

```ruby
ids = DB[:some_table].where { age > 12 }.all

result = event.server.query_members(ids: ids.map { it[:id] }) # => QueriedMembers
```

## Added
`Server#query_members`
`Discordrb::QueriedMembers`

## Changed
Pass the `nonce` and `not_found` data from the chunk to `Server#process_chunk` so the code can tell the difference between query requests, and when we are chunking the entire members list.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -1324,7 +1324,7 @@ module Discordrb
       when :GUILD_MEMBERS_CHUNK
         id = data['guild_id'].to_i
         server = server(id)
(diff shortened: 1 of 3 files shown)


Droid00000:
Should be superseded by #458.

PR #453, opened 2026-05-03
feat: add DAVE E2EE voice support

coderobe:
# Summary

Adds DAVE E2EE voice protocol support through ffi [discord/libdave](https://github.com/discord/libdave) bindings. DAVE support is only enabled when the libdave native libraries are found and can be loaded, so it is not a new hard dependency unless bot operators want to use voice features.

For testing on macos, i provide a libdave homebrew cask so you can `brew tap coderobe/libdave && brew install libdave` and this change should pick it up and enable it automatically.

The path to the libdave shared library can be overridden by setting `DISCORDRB_LIBDAVE_PATH`, and additional libdave-internal debug log verbosity can be tuned with `DISCORDRB_LIBDAVE_LOG_LEVEL` (defaults to `warning`).

---

## Added
`discordrb/voice/dave`: DAVE E2EE protocol support

PR #457, opened 2026-05-18
fix voice packet timing drift causing periodic audio hiccups

coderobe:
# Summary

the send loop was rebasing its timer on the actual clock time after send_audio completed, so encryption and UDP latency accumulated into the inter-packet interval. RTP timestamps still incremented by exactly 20ms per packet, meaning Discord's jitter buffer would gradually drain and produce a periodic audio glitch when playing longer audio files

this switches to a fixed schedule (next_packet_at += IDEAL_LENGTH) so any per-packet send latency is absorbed by the next frame's sleep rather than compounding. Also rebase the schedule on resume after a pause to avoid bursting buffered frames.

While here: fixed `options: null` (NameError) -> `options: ''` in ffmpeg_command, removed a dead clock_gettime call, and marked the now-noop adjust_interval/adjust_offset/adjust_average/length_override attrs as deprecated

---

## Changed
`Discordrb::Voice` voice_bot.rb, encoder.rb

## Deprecated
`Discordrb::Voice#adjust_interval`
`Discordrb::Voice#adjust_offset`
`Discordrb::Voice#adjust_average`
`Discordrb::Voice#length_override`

## Fixed
`Discordrb::Voice`: voice_bot audio playback stutter

### DIFF
--- a/lib/discordrb/voice/encoder.rb
+++ b/lib/discordrb/voice/encoder.rb
@@ -93,7 +93,7 @@ module Discordrb::Voice
 
     private
 
-    def ffmpeg_command(input: '-', options: null)
+    def ffmpeg_command(input: '-', options: '')
       [
         @use_avconv ? 'avconv' : 'ffmpeg',
         '-loglevel', '0',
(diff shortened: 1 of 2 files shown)


coderobe:
without this change, any audio longer than a brief sound effect seems to repeatedly hiccup in semi-regular intervals. tested cherry-picked on top of https://github.com/shardlab/discordrb/pull/453

PR #481, opened 2026-09-22
fix: remove obsolete active developer flag

gitnickolson:
# Summary

This PR removes the active developer flag since the badge does not exist anymore, and the flag is not used anywhere (as far as I could see).

## Removed
- `:active_developer` flag in `Discordrb::Data::User`


Droid00000:
This would be a breaking change.

gitnickolson:
What would be broken? 
@Droid00000

Droid00000:
The flags are used to auto-generate [predicate methods](https://github.com/shardlab/discordrb/blob/05cd95d27c50685e663c1ce12fb7570d7a29a198/lib/discordrb/data/user.rb#L106) such as `User#staff?`, `User#hypesquad?`, etc. Removing the mapping for the active developer flag would in turn remove the `User#active_developer?` flag method start raising `NoMethodError`.

gitnickolson:
I see, meta programming at it's finest!

I may remove the corresponding method calls as well if you don't have any objections.

gitnickolson:
Closing this since we're trying to prevent breaking changes as much as we can.

